PR #92, opened 2021-11-08
Update dependency requirements to be less aggresive

StannisZhou:
Current dependency requirements for common packages (e.g. `numpy`, `scipy`, `matplotlib`) are quite aggressive. This causes some extra steps (e.g. when using on colab). This PR relaxes some of the requirements.

Also gets rid of the warning from dependabot.


NishanthJKumar (approved):
LGTM! Note that I changed my contact email from my Vicarious email (which is now gone) to my MIT CSAIL email in case people do want to contact me (feel free to change this if it's not okay for me to use my MIT email there!)

PR #93, opened 2021-11-08
RCN implementation on a small train and test set

shrinuKushagra:
This PR contains the first implementation of the RCN example using the PGMax package.
The file to run is `examples/rcn/inference_pgmax_small.py`
This code contains implementation, visualization on a small set. Trained with 20 examples and tested on 20 examples.

The inference has been separated from model creation code. 
Saved models are added to `/storage/users/skushagra/pgmax_rcn_artifacts/` . 

Implementation on the full dataset will be implemented in a later PR.


StannisZhou:
Can we get the (standalone) notebook in a separate PR? The notebook should be self-contained (without reference to other files). We can keep this PR for later evaluation on the full test set.

examples/rcn/inference_pgmax_small.py
```diff
@@ -0,0 +1,203 @@
...
+from load_data import get_mnist_data_iters
+
+from pgmax.fg import graph, groups
+
+os.environ["XLA_PYTHON_CLIENT_ALLOCATOR"] = "platform"
+os.environ["XLA_PYTHON_CLIENT_PREALLOCATE"] = "false"
+
+# # 1. Load the data
+
+hps, vps = 11, 11
```
StannisZhou, line 39: Can we make these `12, 12` to match the `science_rcn` repo?

examples/rcn/inference_pgmax_small.py
```diff
@@ -0,0 +1,203 @@
+# ---
```
StannisZhou, line 1: Can we make this into a standalone notebook in its own PR?

examples/rcn/inference_pgmax_small.py
```diff
@@ -0,0 +1,203 @@
...
+
+fg = graph.FactorGraph(variables=variables_all_models)
+for idx in range(edges.shape[0]):
+    edge = edges[idx]
+
+    for e in edge:
+        i1, i2, r = e
+        fg.add_factor(
+            [(idx, i1), (idx, i2)],
+            phis[r],
```
StannisZhou, line 98: We can easily construct these `phis` on demand so probably better to construct them instead of loading them from disk.

examples/rcn/inference_pgmax_small.py
```diff
@@ -0,0 +1,203 @@
...
+
+# -
+
+# # 4. Run inference
+
+# ## 4.1 Helper functins to compute score that will be useful later
+
+
+@jit
+def jax_get_pgmax_score(map_state, bu_msg, frc, hps, vps):
```
StannisZhou, line 113: As we've talked about, score can be easily calculated using the already constructed `evidence_updates` array. No need to go through the following back and forth conversion.

To be precise, you can get the scores of different models by
```
jax.tree_util.tree_multimap(
    lambda evidence, map: jnp.sum(
        evidence[jnp.arange(map.shape[0]), map]
    ), 
    evidence_updates, 
    map_state
)
```

examples/rcn/inference_pgmax_small.py
```diff
@@ -0,0 +1,203 @@
...
+
+        rd = r - hps + (idx - 1) // (2 * vps + 1)
+        cd = c - vps + (idx - 1) % (2 * vps + 1)
+
+        score += bu_msg[f, rd, cd]
+
+    return score
+
+
+def get_pgmax_scores(map_states, bu_msg, frcs, hps, vps, verbose=False):
```
StannisZhou, line 130: See above for simplifications.

examples/rcn/inference_pgmax_small.py
```diff
@@ -0,0 +1,203 @@
...
+# +
+run_bp_fn, _, get_beliefs_fn = graph.BP(fg.bp_state, 30)
+scores = np.zeros((len(test_set), frcs.shape[0]))
+map_states_dict = {}
+
+for test_idx in range(len(test_set)):
+    img = test_set[test_idx][0]
+
+    start = time.time()
+    bu_msg, evidence_updates = initialize_evidences(img, frcs, hps, vps)
```
StannisZhou, line 153: Only need `evidence_updates`. No need to get the `bu_msg` array.

examples/rcn/inference_pgmax_small.py
shrinuKushagra: Computing `phi` involves the following code


    rows = []
    cols = []
    index = 0
    for i in range(1, M):
        r1, c1 = index_to_rc(i, hps, vps)

        r2_min = max(r1 - r, -hps)
        r2_max = min(r1 + r, hps)
        c2_min = max(c1 - r, -vps)
        c2_max = min(c1 + r, vps)

        for r2 in range(r2_min, r2_max + 1):
            for c2 in range(c2_min, c2_max + 1):
                j = rc_to_index(r2, c2, hps, vps)
                rows.append(i)
                cols.append(j)
                index += 1

which is not too complicated. For simplicity of the current notebook; I didnt include it in the example. We can include it if you think it helps the presentation.

NishanthJKumar:
This PR seems stale and redundant given #96 has been merged. Should we close?

StannisZhou:
> This PR seems stale and redundant given #96 has been merged. Should we close?
Good call. Closed!

PR #96, opened 2021-11-19
RCN example

shrinuKushagra:
This PR contains an example implementation of RCN using the pgmax package. 
We load a pre-trained RCN model on a very small subset of mnist (20 examples) and test on a small subset of mnist (20 examples).
The reported accuracy = 0.80.


examples/rcn/load_data.py
```diff
@@ -0,0 +1,78 @@
+# export
+import os
+
+import cv2
```
StannisZhou, line 4: I don't think it's worth it to introduce an additional dependency just for resizing....You can use https://jax.readthedocs.io/en/latest/jax.image.html instead

examples/rcn/load_data.py
```diff
@@ -0,0 +1,78 @@
...
+import matplotlib.pyplot as plt
+import numpy as np
+
+
+# export
+def get_mnist_data_iters(data_dir, train_size, test_size, full_test_set=False, seed=5):
```
StannisZhou, line 10: As discussed, replace this function with something that can download the MNIST datasets. It seems https://pypi.org/project/tensorflow-datasets/ is a standalone package that works well with JAX, so I suggest we add this to our dependency (by updating https://github.com/vicariousinc/PGMax/blob/master/pyproject.toml) and use it to access MNIST

examples/rcn/load_data.py
```diff
@@ -0,0 +1,78 @@
...
+    """
+    Load MNIST data.
+
+    Assumed data directory structure:
```
StannisZhou, line 14: Not sure we want to assume this kind of data directory structure....Better to work with a loaded dataset where things are already organized into arrays.

examples/rcn/load_data.py
```diff
@@ -0,0 +1,78 @@
...
+                continue
+            if num_per_class is None:
+                samples = sorted(os.listdir(cat_path))
+            else:
+                samples = np.random.choice(sorted(os.listdir(cat_path)), num_per_class)
+
+            for fname in samples:
+                filepath = os.path.join(cat_path, fname)
+                # Resize and pad the images to (200, 200)
+                image_arr = cv2.resize(
```
StannisZhou, line 56: Replace with `jax.image.resize` to minimize dependency

examples/rcn/load_data.py
```diff
@@ -0,0 +1,78 @@
+# export
```
StannisZhou, line 1: If we use `tensorflow_datasets` to load MNIST digits, I see no reason for this to be a separate file. We should be able to do data loading and preprocessing in a few lines in the **STANDALONE** notebook

examples/rcn/mnist_small.py
```diff
@@ -0,0 +1,310 @@
...
+import time
+
+import matplotlib.pyplot as plt
+
+# %%
+# %matplotlib inline
+import numpy as np
+from jax import numpy as jnp
+from jax import tree_util
+from load_data import get_mnist_data_iters
```
StannisZhou, line 26: Get rid of this module and make this notebook standalone

pgmax/fg/graph.py
```diff
@@ -48,6 +48,7 @@ class FactorGraph:
         Mapping[Hashable, groups.VariableGroup],
         Sequence[groups.VariableGroup],
         groups.VariableGroup,
+        Mapping[int, groups.VariableGroup],
```
StannisZhou, line 51: This is very hacky....Could you investigate why this is happening and get a principled solution? I just tried and
```
from typing import Hashable
isinstance(2, Hashable)
```
gives me `True`. Also tests are still failing so this is not fixing the issue.

examples/rcn/preproc.py
```diff
@@ -0,0 +1,171 @@
...
+
+    def __init__(self, num_orients=16, filter_scale=4.0, cross_channel_pooling=False):
+        self.num_orients = num_orients
+        self.filter_scale = filter_scale
+        self.cross_channel_pooling = cross_channel_pooling
+        self.suppression_masks = generate_suppression_masks(
+            filter_scale=filter_scale, num_orients=num_orients
+        )
+
+    def fwd_infer(self, img, brightness_diff_threshold=40.0):
```
StannisZhou, line 44: Would be best to extract the relevant bits for this function and get rid of the redundant bits, and also move this into the standalone notebook. Directly copying this bit can be confusing for a tutorial-type example (especially when certain bits seem redundant).

examples/rcn/mnist_small.py
```diff
@@ -0,0 +1,310 @@
...
+
+# %% [markdown]
+# ## 1. Load the data
+#
+#
+# %%
+hps, vps = 12, 12
+train_size = 20
+test_size = 20
+data_dir = "/storage/users/skushagra/MNIST/"
```
StannisZhou, line 42: This is supposed to be included as a tutorial example....People won't have access to `/storage`

examples/rcn/mnist_small.py
```diff
@@ -0,0 +1,310 @@
...
+# -
+
+
+# %% [markdown]
+# ## 2. Load the model
+#
+#
+
+# %%
+directory = f"/storage/users/skushagra/pgmax_rcn_artifacts/model_science_{train_size}_{hps}_{vps}"
```
StannisZhou, line 63: Please include necessary data files inside `examples/example_data`, and structure the notebook in such a way that it can be run anywhere.

examples/rcn/mnist_small.py
```diff
@@ -0,0 +1,310 @@
...
+
+            for index in indices:
+                r1, c1 = index
+                delta_r, delta_c = r1 - hps, c1 - vps
+
+                index = delta_c + vps + (2 * hps + 1) * (delta_r + hps)
+                unary_msg[v, index] = 1
+
+        evidence_updates[idx] = unary_msg
+    return bu_msg, evidence_updates
```
StannisZhou, line 216: This `bu_msg` is no longer used anywhere. Could you please get rid of this?

StannisZhou (changes requested):
Several major issues:

1. We should aim for **A SINGLE NOTEBOOK FILE** (with a few data files in `examples_data`)
2. There seems to be many redundant/random pieces
3. Since this is intended as a tutorial, please make sure this can actually be run by someone. That means:
    - Make sure the data is accessible
    - Make sure you have properly updated the relevant dependicies
    - Make sure you have detailed docstrings/type hints for all your functions, and relevant comments when needed

shrinuKushagra:
Addressed the comments.

However, I could not find a principled solution to the `Hashable int` problem. 
Seems like there might be some issues with the library itself.
https://github.com/python/mypy/issues/2412

pgmax/fg/graph.py
shrinuKushagra: There does not seem to be a principled solution to this problem. Atleast, I could not find one. So have included the hack for now. 
python/mypy#2412

examples/rcn.py
```diff
@@ -0,0 +1,392 @@
...
+        test_size: Desired number of test images.
+    Returns:
+        train_set: A list of length train_size containing images from the MNIST train dataset.
+        train_labels: Corresponding labels for the train images.
+        test_set: A list of length test_size containing images from the MNIST test dataset.
+        test_labels: Corresponding labels for the test images.
+    """
+
+    mnist_train_size = 60000
+    dataset = fetch_openml("mnist_784", as_frame=False, cache=True)
```
StannisZhou, line 59: Would be great if there is a way to get a progress bar here. If that's not possible, maybe also print another message reading `Fetching the data` before this line? Otherwise it's just silent and feels strange....

examples/rcn.py
```diff
@@ -0,0 +1,392 @@
...
+# %%
+data = np.load("example_data/rcn.npz", allow_pickle=True, encoding="latin1")
+frcs, edges, suppression_masks, filters = (
+    data["frcs"],
+    data["edges"],
+    data["suppression_masks"],
+    data["filters"],
+)
+
+M = (2 * hps + 1) * (2 * vps + 1) + 1
```
StannisZhou, line 123: Can we get rid of this `+ 1` as previously discussed? The 0 state does not make sense in this context and does not match the `science_rcn` implementation.

examples/rcn.py
```diff
@@ -0,0 +1,392 @@
...
+    i1, i2, w = e
+    f1, r1, c1 = frc[i1]
+    f2, r2, c2 = frc[i2]
+
+    img[r1, c1] = 255
+    img[r2, c2] = 255
+    plt.text((c1 + c2) // 2, (r1 + r2) // 2, str(w), color="blue")
+    plt.plot([c1, c2], [r1, r2], color="blue", linewidth=0.5)
+
+plt.imshow(img, cmap="gray")
```
StannisZhou, line 143: Screenshot from 2021-11-20 16-11-50
This visualization doesn't look good. Better to get rid of the boundary part, change coloring (I have lots of trouble reading the figure) by just using white background and use something other than blue for the perturb radiuses?

pgmax/fg/graph.py
```diff
@@ -45,7 +45,8 @@ class FactorGraph:
     """
 
     variables: Union[
-        Mapping[Hashable, groups.VariableGroup],
+        Mapping[int, groups.VariableGroup],
```
StannisZhou, line 48: Better to relax it to `Mapping[Any, groups.VariableGroup]` than restricting it to `int` and `str`, since there is no reason the names should be just `int` or `str`. I will also try to look into this

StannisZhou (commented):
Made some initially comments after trying to run it. Will make more later

examples/rcn.py
```diff
@@ -0,0 +1,392 @@
...
+                j = c2 + vps + (2 * hps + 1) * (r2 + hps)
+                rows.append(i)
+                cols.append(j)
+                index += 1
+
+    return np.stack([rows, cols], axis=1)
+
+
+max_perturb_radii = 25
+phis = []
```
StannisZhou, line 204: Change `phis` to more informative names

examples/rcn.py
```diff
@@ -0,0 +1,392 @@
...
+fg = graph.FactorGraph(variables=variables_all_models)
+for idx in range(edges.shape[0]):
+    edge = edges[idx]
+
+    for e in edge:
+        i1, i2, r = e
+        fg.add_factor(
+            [(idx, i1), (idx, i2)],
+            phis[r],
+            np.zeros(phis[r].shape[0]),
```
StannisZhou, line 223: This can be omitted (it would be automatically initialized to all 0 if omitted)

examples/rcn.py
```diff
@@ -0,0 +1,392 @@
...
+        img: The rgb image to compute bottom up messages on (H x W x 3).
+    Returns:
+        bu_msg: An array of shape [16 x H x W] denoting the presence or absence of an oriented line segment at a particular location.
+                The elements of this array belong to the set {+1, -1}.
+    """
+
+    num_orients = 16
+    brightness_diff_threshold = 40.0
+
+    filtered = np.zeros((filters.shape[0],) + img.shape, dtype=np.float32)
```
StannisZhou, line 249: Add a visualization of the filters

examples/rcn.py
```diff
@@ -0,0 +1,392 @@
...
+    pooled_channels = [-np.ones_like(sf) for sf in localized]
+    for i, pc in enumerate(pooled_channels):
+        for channel_offset, factor in pooled_channel_weights:
+            ch = (i + channel_offset) % num_orients
+            pos_chan = localized[ch]
+            if factor != 1:
+                pos_chan[pos_chan > 0] *= factor
+            np.maximum(pc, pos_chan, pc)
+    bu_msg = np.array(pooled_channels)
+    return bu_msg
```
StannisZhou, line 276: Add a visualization of the `bu_msg` we get from running preproc

examples/rcn.py
```diff
@@ -0,0 +1,392 @@
...
+        evidence_updates: A dictionary containing the initial messages to all the variables in the factor graph.
+    """
+
+    bu_msg = get_bu_msg(test_img)
+
+    evidence_updates = {}
+    for idx in range(frcs.shape[0]):
+        frc = frcs[idx]
+
+        unary_msg = -1 + np.zeros((frc.shape[0], M))
```
StannisZhou, line 294: Make sure your `M=625`

examples/rcn.py
```diff
@@ -0,0 +1,392 @@
...
+        for v in range(frc.shape[0]):
+            f, r, c = frc[v, :]
+            evidence = bu_msg[f, r - hps : r + hps + 1, c - hps : c + hps + 1]
+            indices = np.transpose(np.nonzero(evidence > 0))
+
+            for index in indices:
```
StannisZhou, line 300: This for loop looks quite inefficient. How is this different from directly doing `unary_msgs[v] = evidence.ravel()`?

examples/rcn.py
```diff
@@ -0,0 +1,392 @@
...
+    """
+
+    bu_msg = get_bu_msg(test_img)
+
+    evidence_updates = {}
+    for idx in range(frcs.shape[0]):
+        frc = frcs[idx]
+
+        unary_msg = -1 + np.zeros((frc.shape[0], M))
+        for v in range(frc.shape[0]):
```
StannisZhou, line 295: A jax implementation for the whole thing would be
```
@functools.partial(jax.vmap, in_axes=(None, 0), out_axes=0)
def get_evidence(bu_msg, frc):
    ```
    bu_msg: Array of shape (n_features, M, N)
    frc: Array of shape (n_frcs, 3)
    ```
    return jax.lax.dynamic_slice(
        bu_msg[frc[0]], 
        jnp.array([frc[1] - hps, frc[2] - vps]), 
        jnp.array([2 * hps + 1, 2 * vps + 1])
    ).ravel()
```
Since it's vmapping you can directly apply this to the whole `frcs` array so also no more outer for loops.

StannisZhou (commented):
Made more comments

examples/rcn.py
```diff
@@ -0,0 +1,392 @@
...
+            f, r, c = frc[v, :]
+            evidence = bu_msg[f, r - hps : r + hps + 1, c - hps : c + hps + 1]
```
StannisZhou, line 297: You are using `hps` twice here. It looks like a bug. A harmless one, but still a bug.

examples/rcn.py
```diff
@@ -0,0 +1,392 @@
...
+from pgmax.fg import graph, groups
+
+os.environ["XLA_PYTHON_CLIENT_ALLOCATOR"] = "platform"
+os.environ["XLA_PYTHON_CLIENT_PREALLOCATE"] = "false"
+
+# %% [markdown]
+# # 1. Load the data
+# %%
+hps, vps = 12, 12
+train_size = 20
```
StannisZhou, line 42: In my test, on a GPU with 10 GB of memory, with https://github.com/vicariousinc/PGMax/pull/97 you are now able to run inference with 100 templates in one batch. I suggest we change `train_size` to 100 (so that the detections/performances would look better).

examples/rcn.py
shrinuKushagra: Added.

examples/rcn.py
```diff
@@ -0,0 +1,392 @@
...
+    )
+    test_set, test_labels = _sample_data(
+        full_test_set, full_test_labels, test_size // 10
+    )
+
+    return train_set, np.array(train_labels), test_set, np.array(test_labels)
+
+
+# %%
+train_set, train_labels, test_set, test_labels = fetch_mnist_dataset(
```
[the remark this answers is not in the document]
shrinuKushagra: Removed `train_set`.

examples/rcn.py
```diff
@@ -0,0 +1,392 @@
...
+    )
+    full_test_set, full_test_labels = (
+        mnist_images[mnist_train_size:],
+        mnist_labels[mnist_train_size:],
+    )
+
+    train_set, train_labels = _sample_data(
+        full_train_set, full_train_labels, train_size // 10
+    )
+    test_set, test_labels = _sample_data(
```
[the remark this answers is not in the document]
shrinuKushagra: Removed this abstraction.

examples/rcn.py
```diff
@@ -0,0 +1,392 @@
...
+        full_test_set, full_test_labels, test_size // 10
+    )
+
+    return train_set, np.array(train_labels), test_set, np.array(test_labels)
```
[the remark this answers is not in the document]
shrinuKushagra: Removed.

pgmax/fg/graph.py
shrinuKushagra: Changed.

examples/rcn.py
shrinuKushagra: Changed.

examples/rcn.py
shrinuKushagra: Added.

examples/rcn.py
```diff
@@ -0,0 +1,392 @@
...
+# %% [markdown]
+# # 6. Visualize predictions
+
+# %%
+test_idx = 0
+plt.imshow(test_set[test_idx], cmap="gray")
+
+
+# %% [markdown]
+# ## 6.1 Backtrace of some models on this test image
```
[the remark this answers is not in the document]
shrinuKushagra: I changed the visualization to have backtraces of top model for some test images.

Overlaying seemed a bit non-intuitive. Let me know how this looks.

examples/rcn.py
shrinuKushagra: I tried this implementation. Currently, have it commented out.
But when I ran this code, it was slower than the for loop (0.05 sec vs 0.4 sec). Maybe it will be faster when using larger size datasets?

examples/rcn.py
shrinuKushagra: Using `train_size = 100` has the following advantages. 
1. Backtraces might look better. 
2. Higher accuracy.

Disadvantages.
1. A person who does not have a powerful gpu will probably not be able to use it. 
2. Notebook takes a long time to finish on the CPU. For small train set, notebook finishes in a reasonable time even on the CPU. 

If the backtraces look decent on for the smaller case, might be better to use size = 20. Let me know what you think.

StannisZhou: Good point on the requirement of a powerful GPU. Maybe we can give user both options? So load all 100 models, and subset to use only 20 if needed (a simple indexing change), and add a comment detailing the considerations/trade-off.

examples/rcn.py
StannisZhou: Thanks! Looking much better. Can we further zoom in to the actual model and get rid of all the blank space around it (so `img[60:130, 60:160]` or something similar you see fit)

examples/rcn.py
```diff
@@ -0,0 +1,392 @@
...
+        r: Peturb radius
+    Returns:
+        phi: A configuration matrix (shape n X 2) where n is the number of valid configurations.
+    """
+
+    rows = []
+    cols = []
+    index = 0
+    for i in range(M):
+        r1, c1 = -hps + i // (2 * vps + 1), -vps + i % (2 * vps + 1)
```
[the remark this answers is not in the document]
StannisZhou: Reminder that this is still an outstanding comment (and should be an easy one to address). 

Even if the for loop is not that slow, this kind of direct manipulation would be hard to read (especially when there are readily available builtin functions for the same purpose).

For the `min/max` operations you can use array indexing to pick the relevant entries later.

examples/rcn.py
StannisZhou: Reminder

examples/rcn.py
StannisZhou: Reminder

examples/rcn.py
StannisZhou: Might be because no `jit` is used (I didn't include it in the above snippet). Let me test it and see. 

After the change to `ravel` having this for loop in the notebook is not too bad. I'll report back once I play with it a bit.

examples/rcn.py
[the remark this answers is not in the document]
StannisZhou: Screenshot from 2021-11-29 20-01-45
Since we do overlaying all the time not sure how it's non-intuitive. Do you mean the detection doesn't look that good? That is understandable as we are working with a small number of templates.

examples/rcn.py
shrinuKushagra: Changed.

examples/rcn.py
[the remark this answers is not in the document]
shrinuKushagra: Added overlay.

examples/rcn.py
shrinuKushagra: Added support for 100 templates but the default is still 20.
Can switch and wrote comments for explanation.

shrinuKushagra:
All the comments are now addressed. 
Let me know once you run it and other comments you have.

examples/rcn.py
StannisZhou: After adding `jit` timing becomes similar (still slightly slower at 0.08). But it seems to create some issues regarding memory usage so I still went with the CPU version.

StannisZhou:
Want to mention that not all comments are addressed. There were 3 unaddressed comments (with reminders), eg. https://github.com/vicariousinc/PGMax/pull/96#discussion_r754666201

But I addressed them in my commits. Left a few more comments from playing with the notebook. Once these are addressed we can merge the PR.

examples/rcn.py
```diff
@@ -0,0 +1,390 @@
...
+
+            test_set.append(img)
+            test_labels.append(i)
+
+    return test_set, np.array(test_labels)
+
+
+# %%
+test_set, test_labels = fetch_mnist_dataset(test_size)
+train_labels = (
```
StannisZhou, line 93: I tried subsetting a set of trained models, and evaluation breaks completely. In general this seems a very unreliable way to get training labels. Please update `example_data/rcn_100.npz` to include:
1. The 100 training images
2. The labels for the 100 training images
3. The 100 trained RCN models

And load the `train_labels` from `rcn_100.npz`.

examples/rcn.py
```diff
@@ -0,0 +1,390 @@
...
+    np.array([[i] * (train_size // 10) for i in range(10)]).reshape(1, -1).squeeze()
+)
+
+# %% [markdown]
+# # 2. Load the model
+
+# %%
+data = np.load("example_data/rcn_100.npz", allow_pickle=True, encoding="latin1")
+frcs, edges, suppression_masks, filters = (
+    data["frcs"][:train_size],
```
StannisZhou, line 103: Please include additional data so that this kind of subsetting works properly. Right now it seems it would break if I use values other than 20 or 100.

examples/rcn.py
```diff
@@ -0,0 +1,390 @@
...
+    data["edges"][:train_size],
+    data["suppression_masks"],
+    data["filters"],
+)
+
+M = (2 * hps + 1) * (2 * vps + 1)
+
+# %% [markdown]
+# # 3. Visualize loaded model
```
StannisZhou, line 112: For visualizing the model, please also visualize the corresponding training image side by side

examples/rcn.py
```diff
@@ -0,0 +1,390 @@
...
+    Returns:
+        test_set: A list of length test_size containing images from the MNIST test dataset.
+        test_labels: Corresponding labels for the test images.
+    """
+
+    mnist_train_size = 60000
+    num_per_class = test_size // 10
+
+    print("Fetching the MNIST dataset")
+    dataset = fetch_openml("mnist_784", as_frame=False, cache=True)
```
StannisZhou, line 59: From playing with the notebook, the caching does not seem to work. Please investigate and we should make caching work.

examples/rcn.py
```diff
@@ -0,0 +1,390 @@
...
+ax[1].axis("off")
+ax[1].set_title("Max filter response across 16 channels", fontsize=30)
+fig.tight_layout()
+
+
+# %% [markdown]
+# ## 5.2 Run map product inference on all test images
+
+# %%
+def get_evidence(bu_msg: np.ndarray, frc: np.ndarray):
```
StannisZhou, line 305: Return type (I forgot to add)

StannisZhou (commented):
Hopefully the final round of comments

examples/rcn.py
StannisZhou: I now see that we need training images from each of the categories so simple sunsetting won't work....But it would be good in any case to have just one set of models and subset from that. One possibility is to change train size to something like  training image per class and so sunsetting accordingly.

examples/rcn.py
```diff
@@ -0,0 +1,392 @@
...
+# %% [markdown]
+# # 3. Visualize loaded model
+
+# %%
+img = np.ones((200, 200))
+pad = 55
+frc, edge = frcs[4], edges[4]
+fig, ax = plt.subplots(1, 1, figsize=(10, 10))
+for e in edge:
+    i1, i2, w = e
```
swaroopgj, line 122: It might be worth adding some context here in comments or notebook cells that explain what the ws are.

examples/rcn.py
```diff
@@ -0,0 +1,390 @@
...
+def get_evidence(bu_msg: np.ndarray, frc: np.ndarray):
+    """Function to get evidence
+
+    Args:
+        bu_msg: Array of shape (n_features, 200, 200). Contains BU messages
+        frc: Array of shape (n_frcs, 3).
+
+    Returns:
+        evidence; Array of shape (n_frcs, M). Contains evidence
+    """
```
wlehrach, line 316:
```suggestion
    """Returns the evidence (shape (n_frcs, M))

    Args:
        bu_msg: Array of shape (n_features, 200, 200). Contains BU messages
        frc: Array of shape (n_frcs, 3).
    """
```
You can make this concise if you want

> Returns: (or Yields: for generators)
>    Describe the type and semantics of the return value. If the function only returns None, this section is not required. It may also be omitted if the docstring starts with Returns or Yields (e.g. """Returns row from Bigtable as a tuple of strings.""") and the opening sentence is sufficient to describe the return value. 

https://google.github.io/styleguide/pyguide.html

examples/rcn.py
```diff
@@ -0,0 +1,392 @@
...
+    f"example_data/rcn_{train_size}.npz", allow_pickle=True, encoding="latin1"
+)
+frcs, edges, suppression_masks, filters = (
+    data["frcs"][:train_size],
+    data["edges"][:train_size],
+    data["suppression_masks"],
+    data["filters"],
+)
+
+M = (2 * hps + 1) * (2 * vps + 1)
```
swaroopgj, line 111: Needs some explanation, unless the variable is named in a interpretable way.
For this particular variable, better to move where it is first used for better context.

examples/rcn.py
```diff
@@ -0,0 +1,392 @@
...
+# %% [markdown]
+# # 4. Make pgmax graph
+
+# %% [markdown]
+# ## 4.1 Make variables
+
+# %%
+start = time.time()
+assert frcs.shape[0] == edges.shape[0]
+
```
swaroopgj, line 161: I would assign `M` here so that it is clear what it is and why it is defined as such (different categorical values PGM variables can take and how they correspond to pool choices)

examples/rcn.py
```diff
@@ -0,0 +1,392 @@
...
+
+# %% [markdown]
+# ## 4.2 Make factors
+
+# %% [markdown]
+# ### 4.2.1 Pre-compute the valid configs for different perturb radii.
+
+# %%
+def valid_configs(r: int) -> np.ndarray:
```
swaroopgj, line 180: Can you include pool sizes as inputs to this function? 
That makes it clear what is needed to compute these configs.

examples/rcn.py
```diff
@@ -0,0 +1,392 @@
...
+        c2_max = min(c1 + r, 2 * vps)
+        j = np.ravel_multi_index(
+            tuple(np.mgrid[r2_min : r2_max + 1, c2_min : c2_max + 1]), (25, 25)
+        ).ravel()
+        configs.append(np.stack([np.full(j.shape, fill_value=i), j], axis=1))
+
+    return np.concatenate(configs)
+
+
+max_perturb_radius = 25
```
swaroopgj, line 204: I would group this and hps,vps and any other model related params at one place and add a comment stating that.

examples/rcn.py
```diff
@@ -0,0 +1,392 @@
...
+
+# %% [markdown]
+# ### 4.2.2 Make the factor graph
+
+# %%
+start = end
+fg = graph.FactorGraph(variables=variables_all_models)
+for idx in range(edges.shape[0]):
+    edge = edges[idx]
+
```
swaroopgj, line 215: Worth adding a comment on what you are adding here like how you are mapping RCN model edges to factor graph.

examples/rcn.py
```diff
@@ -0,0 +1,392 @@
...
+# %%
+def get_bu_msg(img: np.ndarray) -> np.ndarray:
+    """Computes the bottom-up messages given a test image.
+
+    Args:
+        img: The rgb image to compute bottom up messages on (H x W x 3).
+
+    Returns:
+        An array of shape [16 x H x W] denoting the presence or absence of an oriented line segment at a particular location.
```
swaroopgj, line 241: These are not just oriented but also directional

examples/rcn.py
```diff
@@ -0,0 +1,392 @@
...
+            The elements of this array belong to the set {+1, -1}.
+    """
+
+    num_orients = 16
+    brightness_diff_threshold = 40.0
```
swaroopgj, line 246: Again, these can be grouped and defined along with model params if neededm, but these are not arbitrary as they should match the model you loaded, so read it from the loaded model (filters). BDT can be but not num_orients

examples/rcn.py
```diff
@@ -0,0 +1,392 @@
...
+        kern = filters[i, :, :]
+        filtered[i] = fftconvolve(img, kern, mode="same")
+
+    localized = np.zeros_like(filtered)
+    cross_orient_max = filtered.max(0)
+    filtered[filtered < 0] = 0
+    for i, (layer, suppress_mask) in enumerate(zip(filtered, suppression_masks)):
+        competitor_maxs = maximum_filter(layer, footprint=suppress_mask, mode="nearest")
+        localized[i] = competitor_maxs <= layer
+    localized[cross_orient_max > filtered] = 0
```
swaroopgj, line 259: Add comment explaining what we are doing in this step

examples/rcn.py
```diff
@@ -0,0 +1,392 @@
...
+
+    pooled_channel_weights = [(0, 1), (-1, 1), (1, 1)]
+    pooled_channels = [-np.ones_like(sf) for sf in localized]
+    for i, pc in enumerate(pooled_channels):
+        for channel_offset, factor in pooled_channel_weights:
+            ch = (i + channel_offset) % num_orients
+            pos_chan = localized[ch]
+            if factor != 1:
+                pos_chan[pos_chan > 0] *= factor
+            np.maximum(pc, pos_chan, pc)
```
swaroopgj, line 273: Add comment explaining what we are doing in this step

examples/rcn.py
```diff
@@ -0,0 +1,392 @@
...
+
+    bu_msg = np.array(pooled_channels)
+    bu_msg[bu_msg == 0] = -1
```
swaroopgj, line 276: Add comment explaining what you are doing here and why.

examples/rcn.py
```diff
@@ -0,0 +1,392 @@
...
+# %%
+def get_evidence(bu_msg: np.ndarray, frc: np.ndarray):
+    """Function to get evidence
+
+    Args:
+        bu_msg: Array of shape (n_features, 200, 200). Contains BU messages
+        frc: Array of shape (n_frcs, 3).
+
+    Returns:
+        evidence; Array of shape (n_frcs, M). Contains evidence
```
swaroopgj, line 315:
```suggestion
        evidence: Array of shape (n_frcs, M). Contains evidence
```

examples/rcn.py
```diff
@@ -0,0 +1,392 @@
...
+    img[r_bu_msg[i] > 0] = 0
+
+ax[1].imshow(img, cmap="gray")
+ax[1].axis("off")
+ax[1].set_title("Max filter response across 16 channels", fontsize=30)
+fig.tight_layout()
+
+
+# %% [markdown]
+# ## 5.2 Run map product inference on all test images
```
swaroopgj, line 304: map?

examples/rcn.py
```diff
@@ -0,0 +1,392 @@
...
+        scores[test_idx, ii] = score[ii]
+    end = time.time()
+    print(f"Computing scores took {end-start:.3f} seconds for image {test_idx}.")
+
+
+# %% [markdown]
+# # 6. Compute metrics (accuracy)
+
+# %%
+best_model_idx = np.argmax(scores, axis=1)
```
swaroopgj, line 365: what if we have two classes that score the same?

swaroopgj (commented):
Thanks for adding. I have left comments regarding readability since these are intended for understanding.
If possible I would add a short text at the top to describe RCN model with reference to the paper and the Science implementation available.

shrinuKushagra:
> Want to mention that not all comments are addressed. There were 3 unaddressed comments (with reminders), eg. [#96 (comment)](https://github.com/vicariousinc/PGMax/pull/96#discussion_r754666201)
> 
> But I addressed them in my commits. Left a few more comments from playing with the notebook. Once these are addressed we can merge the PR.

Hi Stannis,
Sorry about this. I went back and checked which of the comments were not addressed. It seems that the following comments were missed.

1. Can we instead use np.unravel_index/ravel_multi_index for these?   -- In the context of the `valid_configs` function.
2. Changing the for loop in valid_config construction 
3 This can be omitted (it would be automatically initialized to all 0 if omitted) -- In the context of `valid_configs`

I went over the `Conversations` tab multiple times and these comments were not there. But looking back at the github emails, I do see that you had made these comments and also reminders for this. The only possible explanation seems that, these comments were present in the `Files` tab and not in the conversations tab. I just realized that sometimes there are comments that only show in the `Files` tab and not in `Conversations` tab. (this is in the context of another pr).

To conclude, I want to apologize for missing them. It was genuine over-sight (from not knowing about a github feature) and I will be more careful in future. Also, thanks for addressing the comments.

examples/rcn.py
```diff
@@ -0,0 +1,390 @@
...
+# Use train_size = 100 if you have a gpu with atleast 8Gbs memory.
+# Recommend that jax is installed with cuda enabled for this option.
+train_size = 100
+test_size = 20
+
+
+def fetch_mnist_dataset(test_size: int, seed: int = 5):
+    """Returns test images sampled randomly from the set of MNIST images.
+    Args:
+        test_size: Desired number of test images.
```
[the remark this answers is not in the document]
shrinuKushagra: Added.

examples/rcn.py
```diff
@@ -0,0 +1,390 @@
...
+# %%
+hps, vps = 12, 12
+
+# Use train_size = 100 if you have a gpu with atleast 8Gbs memory.
+# Recommend that jax is installed with cuda enabled for this option.
+train_size = 100
+test_size = 20
+
+
+def fetch_mnist_dataset(test_size: int, seed: int = 5):
```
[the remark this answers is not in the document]
shrinuKushagra: Added.

examples/rcn.py
shrinuKushagra: Updated. The file is now called `rcn.npz` and train sizes below 100 can be sub-sampled.

examples/rcn.py
shrinuKushagra: Please refer to the reply above.

examples/rcn.py
shrinuKushagra: This seems to be a known issue.
https://github.com/scikit-learn/scikit-learn/issues/18783

Fixed using the suggestions in this thread.

examples/rcn.py
shrinuKushagra: Added.

examples/rcn.py
shrinuKushagra: Changed to this style.

examples/rcn.py
```diff
@@ -0,0 +1,390 @@
...
+
+from pgmax.fg import graph, groups
+
+os.environ["XLA_PYTHON_CLIENT_ALLOCATOR"] = "platform"
+os.environ["XLA_PYTHON_CLIENT_PREALLOCATE"] = "false"
+
+# %% [markdown]
+# # 1. Load the data
+# %%
+hps, vps = 12, 12
```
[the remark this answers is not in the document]
shrinuKushagra: Added.

examples/rcn.py
shrinuKushagra: Added.

examples/rcn.py
shrinuKushagra: Added.

examples/rcn.py
shrinuKushagra: Added.

examples/rcn.py
shrinuKushagra: Added.

examples/rcn.py
shrinuKushagra: Added configs at one place.

examples/rcn.py
shrinuKushagra: Added comment.

examples/rcn.py
shrinuKushagra: Changed the text.

examples/rcn.py
shrinuKushagra: Added configs at one place as suggested above.

examples/rcn.py
shrinuKushagra: Added explanation.

examples/rcn.py
shrinuKushagra: Added explanation.

examples/rcn.py
shrinuKushagra: Added explanation.

examples/rcn.py
shrinuKushagra: Removed.

examples/rcn.py
shrinuKushagra: Expanded on this.

examples/rcn.py
shrinuKushagra: `argmax` will chose the one with smaller value.

examples/rcn.py
```diff
@@ -0,0 +1,438 @@
...
+#     name: python3
+# ---
+
+# %% [markdown]
+# # RCN implementation using the PGMax package
+
+# %% [markdown]
+# RCN is a neuroscience based, data-efficient, interpretable object detection/vision system. It is a probabilistic graphical model (PGM) that trains in minutes and generalizes well (for certain traits). It can reason about, and is robust to, occlusion. It can also imagine and interpret scenes.
+#
+# Please refer to the [Science](https://www.cs.jhu.edu/~ayuille/JHUcourses/ProbabilisticModelsOfVisualCognition2020/Lec21/A%20generative%20vision%20model%20that%20trains%20with%20high%20data%20efficiency%20and%20breaks%20text-based%20CAPTCHAs.pdf) paper for more details. In this notebook, we implement a two-level RCN model on a small subset of the MNIST dataset.
```
swaroopgj, line 22: :roll_eyes: 
Can you use this link please?
https://www.science.org/doi/10.1126/science.aag2612

@lazarox do we have open-access link to share?

swaroopgj: I would also link to the earlier repo https://github.com/vicariousinc/science_rcn since you used that to train the model I assume.

examples/rcn.py
```diff
@@ -0,0 +1,438 @@
...
+    data["frcs"][idxs],
+    data["edges"][idxs],
+    data["suppression_masks"],
+    data["filters"],
+)
+
+hps, vps = 12, 12  # Horizontal and vertical pool sizes respectively for RCN models.
+num_orients = filters.shape[0]
+brightness_diff_threshold = 40.0
+max_perturb_radius = 22
```
swaroopgj, line 131: I would re-group them and explain every variable defined here, with either an inline comment or a comment above it, like you did for pool sizes.
num_orients and bdt are related to preproc to run edge detection.
different from the core PGM model.

swaroopgj: It also won't hurt to expand the pool size to indicate what it exactly is, is it half the window size, is it radius or full size?

examples/rcn.py
swaroopgj: I know but that is wrongly reducing the accuracy.

examples/rcn.py
shrinuKushagra: Changed the link.

examples/rcn.py
```diff
@@ -0,0 +1,438 @@
...
+# %% [markdown]
+# In this notebook, we train an RCN model on a small subset of MNIST (training on 20 images and testing on 20 images). We also have support to train the model on 100 images. Use this option only if you have a gpu with atleast 8Gbs memory. Also, recommend that jax is installed with cuda enabled for training with more than 20 images.
+
+# %%
+train_size = 20
+test_size = 20
+
+
+# %%
+def fetch_mnist_dataset(test_size: int, seed: int = 5) -> Tuple[np.ndarray, np.ndarray]:
```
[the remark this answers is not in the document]
shrinuKushagra: Seems that using `tuple` was enabled in python 3.9?
https://www.python.org/dev/peps/pep-0585/

I was trying to use the suggestion but keep getting `TypeError: 'type' object is not subscriptable`. The project is using python version of `3.7.5`, maybe thats why?

examples/rcn.py
```diff
@@ -0,0 +1,438 @@
...
+
+train_set, train_labels, frcs, edges, suppression_masks, filters = (
+    data["train_set"][idxs, :, :],
+    data["train_labels"][idxs],
+    data["frcs"][idxs],
+    data["edges"][idxs],
+    data["suppression_masks"],
+    data["filters"],
+)
+
```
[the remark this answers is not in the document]
shrinuKushagra: Added detailed description of the parameters.
Moved `max_perturb_radius` to the relevant cell as it not central to the model and only used to save computation.

examples/rcn.py
shrinuKushagra: Done.

examples/rcn.py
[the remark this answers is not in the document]
wlehrach: Ahh - my bad. You need to add:
```
from __future__ import annotations
```
to the top of the file - it'll be default in future python version
You often want that for other reasons, as various other typing things will fail otherwise.

examples/rcn.py
```diff
@@ -0,0 +1,452 @@
...
+from typing import Tuple
+
+import jax
+import matplotlib.pyplot as plt
+import numpy as np
+from jax import numpy as jnp
+from jax import tree_util
+from joblib import Memory
+
+memory = Memory("./example_data/tmp")
```
StannisZhou, line 37: Does this have to be here? If not maybe move it after all the imports

examples/rcn.py
```diff
@@ -0,0 +1,452 @@
...
+# %%
+test_set, test_labels = fetch_mnist_dataset(test_size)
+
+# %% [markdown]
+# # 2. Load the model
+
+# %% [markdown]
+# We load the following variables (that have been pre-computed).
+# - train_set and train_labels - A sample of MNIST train dataset containing 100 train images and their labels.
+# - frcs and edges - Used to represent the learned rcn graphical models.
```
StannisZhou, line 117: Link to the learning module in science RCN repo?

examples/rcn.py
```diff
@@ -0,0 +1,452 @@
...
+
+end = time.time()
+print(f"Creating variables took {end-start:.3f} seconds.")
+
+
+# %% [markdown]
+# ## 4.2 Make factors
+
+# %% [markdown]
+# ### 4.2.1 Pre-compute the valid configs for different perturb radii.
```
StannisZhou, line 229: Radii -> radius for consistency?

examples/rcn.py
```diff
@@ -0,0 +1,452 @@
...
+            tuple(np.mgrid[r2_min : r2_max + 1, c2_min : c2_max + 1]),
+            (2 * hps + 1, 2 * vps + 1),
+        ).ravel()
+        configs.append(np.stack([np.full(j.shape, fill_value=i), j], axis=1))
+
+    return np.concatenate(configs)
+
+
+max_perturb_radius = (
+    22  # The maximum perturb radii for which to pre-compute the valid configs.
```
StannisZhou, line 262: radii -> radius

examples/rcn.py
```diff
@@ -0,0 +1,452 @@
...
+    img[r_bu_msg[i] > 0] = 0
+
+ax[1].imshow(img, cmap="gray")
+ax[1].axis("off")
+ax[1].set_title("Max filter response across 16 channels", fontsize=20)
+fig.tight_layout()
+
+
+# %% [markdown]
+# ## 5.2 Run maximum a-posteriori probability (map) inference on all test images
```
StannisZhou, line 363: I believe the word probability is redundant here

pyproject.toml
```diff
@@ -26,6 +26,7 @@ scipy = "^1.2.3"
 jupytext = "^1.11.3"
 tqdm = "^4.61.0"
 sphinx = {version = "^3", optional = true}
+scikit-learn = "^1.0.1"
```
StannisZhou, line 29: Reminder to also add joblib as part of dependency

examples/rcn.py
StannisZhou: Seems like we are ignoring E402 in flake8 for some reason....Better to move it after all imports

examples/rcn.py
```diff
@@ -0,0 +1,452 @@
...
+pad = 44
+img_idx = 4
+
+model_img = np.ones((200, 200))
+fig, ax = plt.subplots(1, 2, figsize=(20, 10))
+
+
+frc, edge, train_img = frcs[img_idx], edges[img_idx], train_set[img_idx]
+ax[0].imshow(train_img[pad : 200 - pad, pad : 200 - pad], cmap="gray")
+ax[0].set_title("Sample training image", fontsize=20)
```
StannisZhou, line 161: Increase fontsize to 40

examples/rcn.py
```diff
@@ -0,0 +1,452 @@
...
+
+for e in edge:
+    i1, i2, w = e  # The vertices for this edge along with the perturn radius.
+    f1, r1, c1 = frc[i1]
+    f2, r2, c2 = frc[i2]
+
+    model_img[r1, c1] = 0
+    model_img[r2, c2] = 0
+    ax[1].text(
+        (c1 + c2) // 2 - pad, (r1 + r2) // 2 - pad, str(w), color="green", fontsize=15
```
StannisZhou, line 171: Increase fontsize to 25

examples/rcn.py
```diff
@@ -0,0 +1,452 @@
...
+    )
+    ax[1].plot([c1 - pad, c2 - pad], [r1 - pad, r2 - pad], color="green", linewidth=0.5)
+
+ax[1].axis("off")
+ax[1].imshow(model_img[pad : 200 - pad, pad : 200 - pad], cmap="gray")
+ax[1].set_title("Corresponding learned rcn model", fontsize=20)
```
StannisZhou, line 177: Increase fontsize to 40

StannisZhou (commented):
Left some minor comments. LGTM for the most part. Will take a final look after the comments are addressed. Thanks for patiently addressing all the comments!

examples/rcn.py
swaroopgj: My earlier comment was intended for potential typo in usage of "map product inference". 
But you can just use MAP inference here without having to expand.

examples/rcn.py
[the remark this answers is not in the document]
shrinuKushagra: Fixed :D

examples/rcn.py
shrinuKushagra: Moved to after imports.

examples/rcn.py
shrinuKushagra: Done.

examples/rcn.py
shrinuKushagra: Done.

examples/rcn.py
shrinuKushagra: Done.

pyproject.toml
shrinuKushagra: Added.

examples/rcn.py
shrinuKushagra: And added new link where the model is loaded.

examples/rcn.py
shrinuKushagra: If two templates have the same score, I am not sure what we can do in this case. It amounts to an arbitrarily selecting between the two (or more) of them.

examples/rcn.py
shrinuKushagra: Added.

examples/rcn.py
shrinuKushagra: Changed to say MAP inference.

StannisZhou (approved):
LGTM. Excited to have this example ready!

PR #97, opened 2021-11-24
Fix GPU memory leak that came up in RCN example

StannisZhou:
The underlying issue seems to be known yet unresolved (https://github.com/google/jax/issues/282). 

Disabling `jit` and being careful about using jax arrays outside the inference function get rid of the leak. Inference is becoming a bit slower (sometimes as much as 2x slower) due to disabling `jit` for `run_bp`.

Also fixed a mypy issue (relaxed a `Hashable` to `Any`) and reduced precommit autoupdate frequence.

### DIFF
--- a/pgmax/fg/graph.py
+++ b/pgmax/fg/graph.py
@@ -20,6 +20,7 @@ from typing import (
     Sequence,
     Tuple,
     Union,
+    cast,
 )
 
 import jax
@@ -45,7 +46,7 @@ class FactorGraph:
     """
 
     variables: Union[
-        Mapping[Hashable, groups.VariableGroup],
+        Mapping[Any, groups.VariableGroup],
         Sequence[groups.VariableGroup],
         groups.VariableGroup,
     ]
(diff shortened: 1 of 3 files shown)


pgmax/fg/graph.py
```diff
@@ -736,8 +744,12 @@ def __setitem__(
...
             "value",
-            update_evidence(
-                jax.device_put(self.value), {name: jax.device_put(data)}, self.fg_state
+            np.asarray(
+                update_evidence(
+                    jax.device_put(self.value),
+                    {name: jax.device_put(data)},
+                    self.fg_state,
+                ),
             ),
```
shrinuKushagra, line 753: Not expert on jax but for my own clarification. 
Do we not need to clear the variables `self.value` and `data` from jax memory memory?

shrinuKushagra (approved):
Looks good to me. 
Seems like using numpy arrays instead of jax arrays in init functions fixed the error.

pgmax/fg/graph.py
StannisZhou: `self.value` and `data` would stay on CPU. Intermediate objects on GPU will be cleared automatically if you use the `platform` option. 

The leak came from a failure to clear large jit compilation cache (the cache is large because it involves a large constant array, i.e. the wiring). A completely leak-free way is to include wiring as part of the arguments, but in my experiments that leads to slow compiling and hurts performance. Current solution is a bit of a compromise but should suffice for most cases.

PR #98, opened 2021-11-29
Docs update

NishanthJKumar:
Updates documentation and GitHub Actions related to both CI and publishing.

Addresses #28

### DIFF
--- a/.github/workflows/publish.yaml
+++ b/.github/workflows/publish.yaml
@@ -1,6 +1,7 @@
 name: PyPI-publish
 
 on:
+  workflow_dispatch:
   release:
     types: [published]
 
@@ -10,7 +11,6 @@ jobs:
     strategy:
       matrix:
         python-version: [3.7]
-        poetry-version: [1.1.8]
     steps:
       #----------------------------------------------
       #       check-out repo and set-up python
@@ -20,9 +20,9 @@ jobs:
       # Install poetry, then build and publish to PyPI
       #------------------------------------------------
     - name: Build and publish to pypi
-      uses: JRubics/poetry-publish@v1.6
+      uses: JRubics/poetry-publish@v1.9
       with:
         python_version: ${{ matrix.python-version }}
-        poetry_version: ${{ matrix.poetry-version }}
+        poetry_version: "==1.1.8"
         pypi_token: ${{ secrets.PYPI_TOKEN }}
         ignore_dev_requirements: "yes"
--- a/docs/source/contributing.rst
+++ b/docs/source/contributing.rst
@@ -31,5 +31,4 @@ Follow these steps to contribute code:
    ``pytest`` (for unit tests) from within your poetry environment from
    the root of the repository.
 4. Submit a pull request (PR) from the particular branch you've changed
-   on your fork of PGMax. Be sure to follow the below checklist prior to
-   submitting a PR.
+   on your fork of PGMax.
--- a/docs/source/contributing.rst
+++ b/docs/source/contributing.rst
@@ -1,0 +1,35 @@
+Contributing to PGMax
+=====================
+
+PGMax is an open-source library, and as such, we very much value
+contributions from the community!
+
+Contributing code using pull requests
+-------------------------------------
+
+We welcome pull requests, in particular for documentation upgrades,
+additional examples, or those issues marked with `good first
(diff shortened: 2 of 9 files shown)


.github/workflows/publish.yaml
```diff
@@ -1,6 +1,7 @@
 name: PyPI-publish
 
 on:
+  workflow_dispatch:
```
StannisZhou, line 4: Out of curiosity, what is this for?

docs/source/contributing.rst
```diff
@@ -0,0 +1,35 @@
...
+
+We do all of our development using git, so basic knowledge is assumed.
+
+Follow these steps to contribute code:
+
+1. Follow the `Developer Installation
+   Instructions <installation.html#developer>`__ to install a fork of PGMax
+   locally on a new branch.
+2. Develop and implement your changes with your favorite code editor (we
+   recommend `VS Code <https://code.visualstudio.com/>`__)
```
StannisZhou, line 28: Is this necessary :joy:

pgmax/fg/graph.py
```diff
@@ -790,11 +790,13 @@ def BP(bp_state: BPState, num_iters: int) -> Tuple[Callable, Callable, Callable]
         num_iters: Number of belief propagation iterations.
 
     Returns:
-        run_bp: Function for running belief propagation for num_iters.
+        tuple containing
+
+            run_bp(callable): Function for running belief propagation for num_iters.
             Optionally takes as input log_potentials updates, ftov_msgs updates,
```
StannisZhou, line 796: Fix indentation?

pgmax/fg/graph.py
```diff
@@ -790,11 +790,13 @@ def BP(bp_state: BPState, num_iters: int) -> Tuple[Callable, Callable, Callable]
         num_iters: Number of belief propagation iterations.
 
     Returns:
-        run_bp: Function for running belief propagation for num_iters.
+        tuple containing
+
+            run_bp(callable): Function for running belief propagation for num_iters.
```
StannisZhou, line 795: I think we don't need these additional `callable` annotations (to be consistent with internal guidelines).

docs/source/contributing.rst
```diff
@@ -0,0 +1,35 @@
...
+3. Ensure your code still passes formatting checks and unit tests by
+   running ``pre-commit run --all-files`` (for formatting checks) and
+   ``pytest`` (for unit tests) from within your poetry environment from
+   the root of the repository.
+4. Submit a pull request (PR) from the particular branch you've changed
+   on your fork of PGMax. Be sure to follow the below checklist prior to
```
StannisZhou, line 34: `the below checklist` -> a missing checklist?

StannisZhou (approved):
LGTM after some minor comments are addressed!

docs/source/index.rst
```diff
@@ -3,13 +3,20 @@ PGMax Reference Documentation
 PGMax is a library for working with Factor Graphs in `JAX <https://jax.readthedocs.io/en/latest/>`_. It currently provides an interface for specifying factor graphs of any type, as well as an efficient implementation of max-product belief propagation and inference on these graphs.
 
 .. toctree::
-   :maxdepth: 2
+   :maxdepth: 1
```
StannisZhou, line 6: Is this the change that makes the API flatter?

.github/workflows/publish.yaml
NishanthJKumar: this creates the manually-run workflow button!

docs/source/contributing.rst
NishanthJKumar: not really - but they mentioned it in JAX's contribution guide, so I thought I'd do the same lol

docs/source/index.rst
NishanthJKumar: it makes the display on the table-of-contents flatter, but doesn't make the actual docs flatter. Once we get things to build with RTD, we can see if we don't like the hierarchy and flatten it if necessary!

docs/source/contributing.rst
NishanthJKumar: good catch - deleted!

docs/source/contributing.rst
StannisZhou: It stood out to me mainly because I use neovim instead of vs code so at least for me the recommendation is not sincere.... But fine to keep!

PR #99, opened 2021-11-29
fixes bug where wrong conf.py path was specified

NishanthJKumar:
should fix rtd build error


StannisZhou (approved):
LGTM. Thanks for the quick investigation and fixing!

PR #100, opened 2021-11-30
Rtd warning fix

NishanthJKumar:
Should fix warnings and issues with the RTD build!

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -79,7 +79,6 @@ jobs:
       #        test docs build only on PR
       #----------------------------------------------
       - name: Test docs build
-        uses: ammaraskar/sphinx-action@master
-        with:
-          docs-folder: "docs/"
-          build-command: 'make html SPHINXOPTS="-W"'
+        run: |
+          cd docs
+          poetry run make html SPHINXOPTS="-W"'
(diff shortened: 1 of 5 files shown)


StannisZhou:
Could you also update https://github.com/vicariousinc/PGMax/blob/1f9633b81e4a1e2d9383d4c798c9a1a1552f4947/.github/workflows/ci.yaml#L17 so that tests are also run for python 3.8?

Also, should we change this fail on warning behavior https://github.com/vicariousinc/PGMax/blob/1f9633b81e4a1e2d9383d4c798c9a1a1552f4947/.readthedocs.yaml#L11 so that CI blocks when docs building actually fails (not sure if this is the argument controlling this behavior)?

NishanthJKumar:
So for the first point, I'm not sure if we want to test both 3.7 and 3.8 (it'll double the time taken for running CI). I'm happy to put it in if you'd like, but maybe we can just test 3.8 (because we'd expect it to work on 3.7 if it works on 3.8)?

For the second part, this flag is saying that if there are any warnings, then RTD will fail (if there are errors, it will also fail by default). I'll update CI to make it fail on warnings!

StannisZhou (approved):
LGTM

OK let's keep CI testing for only python 3.7 for now. But it would be good if we at least do one test somewhere to make sure all tests pass for python 3.8. Maybe you can manually kick off a run on a branch of yours with python 3.8? Just want to make sure nothing breaks unexpectedly.

NishanthJKumar:
Sounds good! I ran tests locally with a python 3.8 env and everything worked!

Issue #115, opened 2022-02-14
Incomplete documentation for add_factor


nathanielvirgo:
The documentation for `add_factor` currently says:

> log_potentials – Optional array of shape (num_val_configs,) or (num_factors, num_val_configs). If specified, it contains the log of the potential value for every possible configuration. If none, it is assumed the log potential is uniform 0 and such an array is automatically initialized.

However, in order to use it one would need to know the order in which the factors should be specified. Could this be added to the documentation?

StannisZhou:
Thanks for bringing this up! This is actually a bug in the documentation. `log_potentials` in this case should always be an array of shape `(num_val_configs,)`, and `add_factor` should always be used to add one factor at a time.

This bug in the documentation was a result of copying the docstring for `log_potentials` from `EnumerationFactorGroup`. I will open a PR to fix this.

See https://github.com/vicariousinc/PGMax/blob/2d86b75169bf45f6e0b5d7b5d1ab564d435df1bc/examples/rbm.py#L82 and Sections 2 and 3 of our [blog post](https://www.vicarious.com/posts/pgmax-factor-graphs-for-discrete-probabilistic-graphical-models-and-loopy-belief-propagation-in-jax/) for an example on how to use `add_factor`. `factor_configs` and `log_potentials` should be a literal translation of factor definitions (see for example Tables 1 and 2 in the blog post).

nathanielvirgo:
Thank you for working on it! Note that we'll still need to know which element of the array of shape `(num_val_configs,)` should correspond to which outcome. The examples in the blog post don't help so much with that, because arrays containing the correct data are just assumed to exist already.

StannisZhou:
`factor_configs` specifies the list of valid configurations. `log_potentials` is assumed to contain one element for each valid configuration in `factor_configs` and the order is assumed to match.

If the arrays containing the correct data don't already exist, I assume you should be able to create the necessary arrays in the right format easily (e.g. maybe you allow all possible combinations in which case you can use `itertools.product` to create the necessary `factor_configs` and specify `log_potentials` accordingly).

If you are not able to create the arrays then that seems to suggest the factor definitions are unclear?

nathanielvirgo:
The values to use for each factor are perfectly clear in my mind, that isn't the issue at all. I'm actually just trying to implement something simple like the sprinkler example, because that's a more useful starting point for me than the complex machine learning examples in the blog post.

[...]

PR #116, opened 2022-02-14
Get rid of redundant array shape for log_potentials

StannisZhou:
Resolves https://github.com/vicariousinc/PGMax/issues/115


NishanthJKumar (approved):
LGTM! Thanks for doing this!

PR #122, opened 2022-03-02
Passing messages from OR factors to variables

antoine-dedieu:
We add a specialized`pass_OR_fac_to_var_messages` which passes messages from OR Factors to Variables.
We compare it with the old enumeration approach via an unit test


pgmax/bp/infer.py
```diff
@@ -97,7 +97,112 @@ def pass_fac_to_var_messages(
...
+
+    top_tof_msgs_pos = jnp.maximum(0.0, top_tof_msgs)
+    sum_top_tof_msgs_pos_inc = (
+        jnp.full(shape=num_factors, fill_value=0.0)
+        .at[factor_indices]
+        .add(top_tof_msgs_pos)
+    )
+
+    # Outgoing messages to top variables
+    top_msgs = jnp.minimum(
```
StannisZhou, line 169: I'm getting confused about this update. Shouldn't the update be (for non-firstargmaxes):
```
top_msgs_to1 = bottom_tof_msgs[factor_indices] + sum_top_tof_msgs_pos_inc[factor_indices] - top_tof_msgs_pos # The given parent is 1, child has to be 1, other parents can take whatever value they want to take
top_msgs_to0 = jnp.maximum(
	0.0, # The given parent is 0, child is 0, all other parents are 0
	bottom_tof_msgs[factor_indices] + sum_top_tof_msgs_pos_inc[factor_indices] - top_tof_msgs_pos + jnp.minimum(0.0, top_tof_msgs[first_argmaxes][factor_indices]) # The given parent is 0, child is 1, at least one other parent needs to be 1
)
top_msgs = top_msgs_to1 - top_msgs_to0
```
Are these equivalent?

StannisZhou: Actually they are equivalent :+1: Might be better to include a bit of comments here (e.g. just copying the above more naive implementation as comments)

pgmax/bp/infer.py
```diff
@@ -97,7 +97,112 @@ def pass_fac_to_var_messages(
...
+        (1 - has_single_parents) * top_msgs[first_argmaxes]
+        + has_single_parents * bottom_tof_msgs
+    )
+
+    # Outgoing messages to bottom variables
+    bottom_msgs = sum_top_tof_msgs_pos_inc + jnp.minimum(
+        0.0, top_tof_msgs[first_argmaxes]
+    )
+
+    bottom_tov0_msgs = jnp.minimum(0.0, -bottom_msgs)
```
StannisZhou, line 196: After each update we have a `normalize_and_clip` step, which would be doing what you are doing here. So no need to do this twice. Just doing
```
ftov_msgs = jnp.zeros_like(vtof_msgs)
ftov_msgs = ftov_msgs.at[children_states[..., 1] + 1].set(bottom_msgs)
ftov_msgs = ftov_msgs.at[parents_states[..., 1] + 1].set(top_msgs)
```
would be good enough.

tests/bp/test_infer.py
```diff
@@ -0,0 +1,119 @@
+from itertools import product
+
+import numpy as np
+
+from pgmax.bp import infer
+
+print("We assume that factor_id is increasing and take values in 0,..., n_factors")
```
StannisZhou, line 7: We should be able to incorporate this assumption when we compile various arrays. Just need to be a bit careful about how to structure the interface but the assumption should be fine

tests/bp/test_infer.py
```diff
@@ -0,0 +1,119 @@
...
+    Compares the specialized pass_OR_fac_to_var_messages implementation which passes messages from OR Factors to Variables
+    to the alternative pass_fac_to_var_messages which enumerates all the possible configurations.
+    """
+
+    for idx in range(10):
+        np.random.seed(idx)
+
+        # Define OR factor and incoming messages
+        num_factors = np.random.randint(3, 6)
+        num_parents = np.random.randint(1, 6, num_factors)
```
StannisZhou, line 21: Should not support single parent case

tests/bp/test_infer.py
```diff
@@ -0,0 +1,119 @@
+from itertools import product
+
+import numpy as np
+
+from pgmax.bp import infer
+
+print("We assume that factor_id is increasing and take values in 0,..., n_factors")
+
+
+def test_pass_fac_to_var_messages():
```
StannisZhou, line 10: Fine to leave it as is for now. Once we added proper interface for OR factors we should reimplement this unit test using automated derived inference functions from a factor graph specified through the interface.

StannisZhou (commented):
First round of comments. Some main issues:

1. Look into adding support for different temperatures
2. Improve naming. Might be better to stick to parents/children instead of top/bottom, and explicitly use positive/incoming instead of just pos/inc.
3. Move the function for computing maxes/argmaxes out as a reusable, standalone function.
4. Should not support single parent case
5. In general more comments would help

pgmax/bp/infer.py
```diff
@@ -97,7 +97,112 @@ def pass_fac_to_var_messages(
...
+    top_msgs = top_msgs.at[first_argmaxes].set(
+        jnp.minimum(
+            bottom_tof_msgs
+            + sum_top_tof_msgs_pos_inc
+            - top_tof_msgs_pos[first_argmaxes],
+            jnp.maximum(0.0, -top_tof_msgs[second_argmaxes]),
+        )
+    )
+
+    # Special case for factor with single parents
```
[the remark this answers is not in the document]
antoine-dedieu: I am not sure. This trivial case can be useful to debug  some complex implementations.

pgmax/bp/infer.py
```diff
@@ -97,7 +97,112 @@ def pass_fac_to_var_messages(
...
+    """Passes messages from OR Factors to Variables.
+
+    Args:
+        vtof_msgs: Array of shape (num_edge_states,).
+            This holds all the flattened (binary) variables to factor messages.
+        parents_states: Array of shape (num_parents, 2)
+            parents_states[ii, 0] contains the global factor index
+            parents_states[ii, 1] contains the message index of the parent variable's state 0.
+            The parent variable's state 1 is parents_states[ii, 2] + 1
+        children_states: Array of shape (num_factors, 2)
```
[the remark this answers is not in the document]
antoine-dedieu: I don't think there should be such possibility. Still for consistency with the `parents_states` array, shouldn't we keep it that way?

pgmax/bp/infer.py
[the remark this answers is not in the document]
StannisZhou: Not arguing with it's usefulness in debugging. I'm only saying that if the user tries to specify such a factor it seems to me the sensible behavior is to raise an error instead of having this block of code.

pgmax/bp/infer.py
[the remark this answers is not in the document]
StannisZhou: I think it would be better to avoid unnecessary redundancy. Also there's no obvious reason for having the parents consistent with the children, as they are two different kinds of variables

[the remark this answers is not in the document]
antoine-dedieu: fine with me

pgmax/bp/infer.py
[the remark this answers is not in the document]
antoine-dedieu: Again I think we should support it. Let's discuss this point today

antoine-dedieu:
@StannisZhou thanks for your comments. The main points to discuss are
1. can we support any temperature (ie derive simple expression)?
2. should we support the single parent case -- I think we should
3. what do we think about attaching the supplementary of the PMP Neurips paper for proper derivation of the equations?

pgmax/bp/infer.py
```diff
@@ -97,7 +97,112 @@ def pass_fac_to_var_messages(
                 )
             )
         )
+    return ftov_msgs
+
+
+@jax.jit
+def pass_OR_fac_to_var_messages(
```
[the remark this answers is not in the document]
StannisZhou: Yes we can discuss and think about it a bit. I think it should be doable...

pgmax/bp/bp_utils.py
```diff
@@ -50,3 +51,37 @@ def get_max(data, start_index, segment_length):
...
+
+@functools.partial(jax.jit, static_argnames="num_labels")
+def get_maxes_and_argmaxes(
+    data: jnp.array, labels: jnp.array, num_labels: int
+) -> Tuple[jnp.ndarray, jnp.ndarray]:
+    """
+    Given a flattened sequence of elements and their corresponding labels,
+    returns the maxes and argmaxes of each label.
+
+    Args:
```
StannisZhou, line 64: Missing docstring for `num_labels`

pgmax/bp/bp_utils.py
```diff
@@ -50,3 +51,37 @@ def get_max(data, start_index, segment_length):
...
+        data: Array of shape (a_len,) where a_len is an arbitrary integer.
+        labels: Label array of shape (a_len,), assigning a label to each entry.
+            Labels must be 0,..., num_labels - 1.
+
+    Returns:
+        Maxes and argmaxes arrays
+    """
+    assert (
```
StannisZhou, line 72: Raise an error instead of using assert

pgmax/bp/bp_utils.py
```diff
@@ -50,3 +51,37 @@ def get_max(data, start_index, segment_length):
...
+        data.shape[0] == labels.shape[0]
+    ), "Data and labels arrays must have the same size!"
+    num_obs = data.shape[0]
+
+    maxes = jnp.full(shape=num_labels, fill_value=-jnp.inf).at[labels].max(data)
```
StannisZhou, line 77: Also use `NEG_INF` for consistency?

pgmax/bp/infer.py
```diff
@@ -97,7 +97,91 @@ def pass_fac_to_var_messages(
...
+    )
+
+    # Special case: factors with a single parent
+    has_single_parents = (first_parents_argmaxes == second_parents_argmaxes).astype(
+        jnp.float32
+    )
+    parents_msgs = parents_msgs.at[first_parents_argmaxes].set(
+        (1 - has_single_parents) * parents_msgs[first_parents_argmaxes]
+        + has_single_parents * children_tof_msgs
+    )
```
StannisZhou, line 175:
```suggestion
    # Special case: factors with a single parent
    parents_msgs = parents_msgs.at[first_parents_argmaxes].set(
    	jnp.where(first_parents_argmaxes == second_parents_argmaxes),
    	children_tof_msgs,
    	parents_msgs[first_parents_argmaxes]
    )
```

tests/bp/test_infer.py
StannisZhou: Synced offline. Let's support it

pgmax/bp/infer.py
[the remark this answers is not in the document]
StannisZhou: Let's support it but make the code more compact. In general it's better to use `jnp.where` instead of `astype(float)` and multiplication.

antoine-dedieu:
@StannisZhou I have started a draftr or the `ORFactor` and the `ORFactorGroup`. Let me know what you think and we can take it from there.
Note that we can implement this as follow:
```
parents_variables = groups.NDVariableArray(num_states=2, shape=(5,))
children_variable = groups.NDVariableArray(num_states=2, shape=(2,))
fg = graph.FactorGraph(
    variables=dict(parents=parents_variables, children=children_variable),
)
parents_names_for_factors = [
    [("parents", 0), ("parents", 1), ("parents", 2)],
    [("parents", 3), ("parents", 4)]
]
children_names_for_factors = [("children", 0), ("children", 1)]

fg.add_factor_group(
    factory=groups.ORFactorGroup,
    parents_names_for_factors=parents_names_for_factors,
    children_names_for_factors=children_names_for_factors,
)
```

pgmax/fg/nodes.py
```diff
@@ -174,3 +174,120 @@ def compile_wiring(
...
+
+    @utils.cached_property
+    def edges_num_states(self) -> np.ndarray:
+        """Number of states for the variables connected to each edge
+
+        Returns:
+            Array of shape (num_edges,)
+            Number of states for the variables connected to each edge
+        """
+        return np.array([2] * (len(self.parents_variables) + 1))
```
StannisZhou, line 258:
```suggestion
        return np.full((len(self.parents_variables) + 1,), fill_value=2, dtype=int)
```

pgmax/fg/nodes.py
```diff
@@ -174,3 +174,120 @@ def compile_wiring(
...
+        Args:
+            vars_to_starts: A dictionary that maps variables to their global starting indices
+                For an n-state variable, a global start index of m means the global indices
+                of its n variable states are m, m + 1, ..., m + n - 1
+
+        Returns:
+            Enumeration wiring for the enumeration factor
+        """
+        num_parents = len(self.parents_variables)
+        parents_states = np.vstack(
```
StannisZhou, line 272: Specify dtype to be int?

pgmax/fg/nodes.py
```diff
@@ -174,3 +174,120 @@ def compile_wiring(
...
+    Raises:
+        ValueError: If:
+            (1) The parents variables are not all binary
+            (2) The child variable is not binary
+            (3) The list of parent variables is empty
+    """
+
+    parents_variables: Tuple[Variable, ...]
+    child_variable: Variable
+    log_potentials: None
```
StannisZhou, line 234: This looks redundant (since there is not need to be consistent with anything here)...

pgmax/fg/graph.py
```diff
@@ -180,31 +183,66 @@ def _register_factor_group(
...
+            return None
+
+        enum_wirings = [
+            factor.compile_wiring(self._vars_to_starts)
+            for factor in self.factors
+            if isinstance(factor, nodes.EnumerationFactor)
         ]
-        wiring = fg_utils.concatenate_enumeration_wirings(wirings)
-        return wiring
+        enum_wiring = fg_utils.concatenate_enumeration_wirings(enum_wirings)
```
StannisZhou, line 219: You need to be careful about this concatenation here. Since there might be both enumeration factors and OR factors present, you might no longer be able to rely on simple offsets to do this concatenation.

pgmax/fg/graph.py
```diff
@@ -180,31 +183,66 @@ def _register_factor_group(
...
+                isinstance(factor_group, nodes.EnumerationFactor)
+                for factor_group in self.factor_groups
+            ]
+        ):
+            return None
+
+        # or_wirings = [
+        #     factor.compile_wiring(self._vars_to_starts) for factor in self.factors if isinstance(factor, nodes.ORFactor)
+        # ]
+        # or_wiring = fg_utils.concatenate_or_wirings(or_wirings)
```
StannisZhou, line 243: Similar comment as above. Might need to be careful about how concatenation is done.

pgmax/fg/graph.py
```diff
@@ -249,7 +296,8 @@ def fg_state(self) -> FactorGraphState:
             factor_to_potentials_starts=copy.copy(self._factor_to_potentials_starts),
             factor_to_msgs_starts=copy.copy(self._factor_to_msgs_starts),
             log_potentials=self.log_potentials,
-            wiring=self.wiring,
+            enum_wiring=self.enum_wiring,
```
StannisZhou, line 299: I still think it's best to organize this inside some containers (e.g. a dictionary). This way, in the future if we add more types of factors we don't need to touch this line again (a sign of good abstraction/organization).

StannisZhou (commented):
Left some comments. Will look at the message updates with different temperature part more carefully later.

One high-level comment is, one way to see how good our abstractions are is to think about how painful it would be in the future to add a new type of factor. This PR is definitely a bit painful. It would be helpful to make sure after this PR it would be much less painful. And important things to consider include minimize the number of places we need to touch when adding new factors in the future (e.g. organizing wirings in a dictionary instead of separate enumerations), and isolate places that need updating to be non-disruptive to other parts of the codebase.

pgmax/fg/nodes.py
antoine-dedieu: this is needed for the `factor_group_log_potentials` method of the `FactorGroup` class

pgmax/fg/fg_utils.py
```diff
@@ -1,50 +1,91 @@
...
 import numpy as np
 
-from pgmax.fg.nodes import EnumerationWiring
+from pgmax.fg.nodes import EnumerationWiring, ORWiring
 
 
 def concatenate_enumeration_wirings(
-    wirings: Sequence[EnumerationWiring],
-) -> EnumerationWiring:
+    enum_wirings: Sequence[EnumerationWiring], start_edge_states: np.ndarray
```
StannisZhou, line 11: We can go with this approach and get to a complete, runnable implementation first, but this way of doing things might get annoying quickly.

If you just aggregate different types of factors into different containers (e.g. one list containing all enumeration factors, and another list containing all OR factors), you can then have one global offset which you apply to all factors. This also decouples things as much as possible.

antoine-dedieu:
@StannisZhou the unit test now builds 2 factors graphs with Enumeration vs OR factors and compares both general vs specialized factor to variables message updates.

pgmax/fg/groups.py
```diff
@@ -1006,3 +1012,66 @@ def unflatten(
             )
 
         return data
+
+
+@dataclass(frozen=True, eq=False)
+class ORFactorGroup(FactorGroup):
```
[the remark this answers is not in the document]
StannisZhou: One possibility is for this to be a group of OR factors with the same number of parents. In that case we also need to implement the `flatten` and `unflatten` functions accordingly.

pgmax/fg/graph.py
```diff
@@ -121,6 +123,7 @@ def add_factor(
                 If None, it is assumed the log potential is uniform 0 and such an array is automatically
                 initialized.
         """
+        # TODO: change
```
StannisZhou, line 126: We should indeed support adding individual OR factors. One possibility is to change this function. Another is to have a new function which takes a factory function and other arguments, similar to the current `add_factor_group` function, just for adding different kinds of specialized factors.

pgmax/fg/graph.py
```diff
@@ -180,31 +187,68 @@ def _register_factor_group(
...
             self._named_factor_groups[name] = factor_group
 
+        if isinstance(
+            factor_group, (groups.EnumerationFactorGroup, groups.PairwiseFactorGroup)
+        ):
+            self._total_factor_num_configs += factor_group_num_configs
+
     @cached_property
-    def wiring(self) -> nodes.EnumerationWiring:
+    def wiring(self) -> GraphWiring:
```
StannisZhou, line 208: This function should be updated in such a way that adding new factors should not involve updating this function.

A possible way of doing it:
- Organize all the factors according to types into a dictionary, which maps from factor types to the list of factors of that type.
- Inside this function, simply looping through the different factor types, call compile wiring on each factor, and call the appropriate concatenation function (probably also organized into a dictionary somewhere) to get the final wiring. The final wiring is also organized into a dictionary.

Additionally, we should minimize interdependency between different types of factors. Concretely, this means we assume factors are ordered according to their types (e.g. all the enumeration factors first, then all the OR factors), and the current `enum_factor_to_msgs_starts` and `or_factor_to_msgs_starts` should be replaced by single numbers (0 and a single offset respectively).

pgmax/fg/graph.py
```diff
@@ -216,15 +260,32 @@ def log_potentials(self) -> np.ndarray:
             A jnp array representing the log of the potential function for each
                 valid configuration
         """
+        if not np.any(
```
StannisZhou, line 263: Instead of doing this kind of `if`, I think again it's better to organize things into dictionaries mapping from factor types to elements, and looping through dictionary elements and return a dictionary.

In this case, one dictionary element would be all enumeration factors, and the original function can be applied without change. The other would be all the OR factors, for which we just return None.

pgmax/fg/graph.py
```diff
@@ -216,15 +260,32 @@ def log_potentials(self) -> np.ndarray:
...
+                if isinstance(
+                    factor_group,
+                    (groups.EnumerationFactorGroup, groups.PairwiseFactorGroup),
+                )
             ]
         )
 
     @cached_property
-    def factors(self) -> Tuple[nodes.EnumerationFactor, ...]:
+    def factors(
```
StannisZhou, line 286: Related to the above comments, instead of returning a single tuple we should return a dictionary, with one element per factor type.

pgmax/fg/graph.py
```diff
@@ -304,6 +369,38 @@ def __post_init__(self):
                 object.__setattr__(self, field, MappingProxyType(getattr(self, field)))
 
 
+@dataclass(frozen=True, eq=False)
+class GraphWiring:
```
StannisZhou, line 373: A more generic structure (e.g. a dictionary) might be better than a new dedicated class (which we have to update every time we add a new factor type).

pgmax/fg/graph.py
```diff
@@ -807,9 +925,14 @@ def BP(
         \tget_beliefs: Function to calculate beliefs from BPArrays.\n
     """
     max_msg_size = int(np.max(bp_state.fg_state.wiring.edges_num_states))
-    num_val_configs = (
-        int(bp_state.fg_state.wiring.factor_configs_edge_states[-1, 0]) + 1
-    )
+
+    if bp_state.fg_state.wiring.enum_wiring is not None:
```
StannisZhou, line 929: Again, instead of having these kind of if/else, we should probably do a simple loop on a generic structure (e.g. a dictionary, one element for each factor type) inside this function. 

We should make enough abstractions that adding new factor types should not involve updating this function.

StannisZhou (commented):
Left some more comments.

The basic functionality is there. The main issue at this moment is that to add a new type of factor we need to touch many many places (and most places in a non-trivial way so unless someone spends time to completely understand all the details of the package there is no way they can add a new type of factor).

We should minimize the number of places we need to touch. Some suggestions include:
1. Introduce base classes for factors and factor wirings, and unify interfaces when possible (e.g. not having separate parent and children variables, but just have one list of variables where the first or last is the child).
2. Introduce enough abstractions (e.g. structuring things as dictionaries with one element per factor type, and do various things by looping through dictionary elements). There is a simple test to see whether we have enough abstractions: we should NOT need to touch `fg/graph.py` when we need to add a new factor type in the future.
3. Reorganize the repo to consolidate pieces related to a particular factor type. We can introduce a new top-level `factors` module, with two submodules `enumeration.py` and `logical.py`. We can define base classes for factors and wirings inside `fg.nodes`, and move individual factor definitions to the submodules under `factors`. We can similarly move the wiring concatenation functions and the factor to variable message updates functions into these submodules. The end result should be that to add a new type of factor we just need to specify these pieces within the corresponding submodule, and update one other place for a global registration of factor types. Adding a new type of factor should not involving changing anything outside the `factors` module.

pgmax/fg/nodes.py
```diff
@@ -57,6 +57,52 @@ def tree_unflatten(cls, aux_data, children):
...
+        return jax.tree_util.tree_flatten(asdict(self))
+
+    @classmethod
+    def tree_unflatten(cls, aux_data, children):
+        return cls(**aux_data.unflatten(children))
+
+
+@jax.tree_util.register_pytree_node_class
+@dataclass(frozen=True, eq=False)
+class ORWiring(LogicalWiring):
```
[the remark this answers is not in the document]
antoine-dedieu: Currently an ORFactor must have an ORWiring. Similarly an ANDFactor will have an ANDWiring
We can discuss this.

pgmax/fg/graph.py
antoine-dedieu: Not here: we need a flattened list of factors because we will have a flattened message to factors

pgmax/fg/graph.py
antoine-dedieu: Agreed

pgmax/fg/graph.py
StannisZhou: I don't think these are contradictory. We basically cut the flat messages into a few sections, one per factor type. Each factor type has access to start and end indices and does message updates using only that section of the messages

pgmax/fg/graph.py
antoine-dedieu: We should discuss your last point. Currently factors and their dependencies (ie `_factor_to_msgs_starts` etc...) organized following the order in which they were added.
We can leave it this way. 
What you suggest is to add the factors then reorganize them, which may create some overhead.

StannisZhou: This is related to the comment for replacing the current tuple of factors with a dictionary. The only change would be to aggregate factors into two separate lists (according to their types) instead of one and I think there's minimal overhead

pgmax/factors/enumeration.py
```diff
@@ -0,0 +1,144 @@
...
+        factor_configs_edge_states: Array of shape (num_factor_configs, 2)
+            factor_configs_edge_states[ii] contains a pair of global enumeration factor_config and global edge_state indices
+            factor_configs_edge_states[ii, 0] contains the global EnumerationFactor config index,
+            which takes into account all the EnumerationFactors and LogicalFactors.
+            factor_configs_edge_states[ii, 1] contains the corresponding global edge_state index
+            which takes into account all the EnumerationFactors.
+    """
+
+    edges_num_states: Union[np.ndarray, jnp.ndarray]
+    var_states_for_edges: Union[np.ndarray, jnp.ndarray]
```
StannisZhou, line 33: These seem to be already specified in the parent class?

pgmax/factors/enumeration.py
```diff
@@ -0,0 +1,144 @@
...
+    Raises:
+        ValueError: If:
+            (1) The dtype of the configs array is not int
+            (2) The dtype of the potential array is not float
+            (3) Configs does not have the correct shape
+            (4) The potential array does not have the correct shape
+            (5) The configs array contains invalid values
+    """
+
+    variables: Tuple[nodes.Variable, ...]
```
StannisZhou, line 57: This is already specified in the parent class

pgmax/factors/logical.py
```diff
@@ -0,0 +1,131 @@
...
+            The last variable is assumed to be the child one.
+        logical_type: The logical condition supported by the factor
+
+    Raises:
+        ValueError: If:
+            (1) The variables are not all binary
+            (2) There are less than 2 variables
+    """
+
+    variables: Tuple[nodes.Variable, ...]
```
StannisZhou, line 58: Already specified in parent class

pgmax/factors/logical.py
```diff
@@ -0,0 +1,131 @@
...
+    logical_type: str
+    log_potentials: None
```
StannisZhou, line 60: Having this here means users can actually overwrite it with a non-None value. Replacing with `field(init=False, default=None)` would be better.

pgmax/fg/groups.py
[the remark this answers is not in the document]
antoine-dedieu: We currently add factor using `FactorGroups` so I would suggest we keep a `LogicalFactorGroup`

pgmax/fg/nodes.py
```diff
@@ -57,6 +57,52 @@ def tree_unflatten(cls, aux_data, children):
         return cls(**aux_data.unflatten(children))
 
 
+@jax.tree_util.register_pytree_node_class
+@dataclass(frozen=True, eq=False)
+class LogicalWiring:
```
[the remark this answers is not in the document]
antoine-dedieu: agreed

pgmax/fg/graph.py
```diff
@@ -235,21 +282,26 @@ def factor_groups(self) -> Tuple[groups.FactorGroup, ...]:
 
     @cached_property
     def fg_state(self) -> FactorGraphState:
-        """Current factor graph state given the added factors."""
+        """Current factor graph state given the added factors.
+        Also computes the factor_to_msgs_starts and the log_potentials."""
+        # Preliminary computations
+        log_potentials = self.log_potentials
+        wiring = self.wiring
```
StannisZhou, line 289: This does not seem to be changing anything. Better to get rid of it to avoid unnecessary diffs.

pgmax/fg/graph.py
StannisZhou: Get rid of this class. Not necessary

pgmax/fg/graph.py
```diff
@@ -432,12 +526,23 @@ def __getitem__(self, name: Any) -> np.ndarray:
 
         if name in self.fg_state.named_factor_groups:
             factor_group = self.fg_state.named_factor_groups[name]
+            if not isinstance(
```
StannisZhou, line 529: This is again a factor type dependent special processing. This means when adding new factors we potentially need to touch this file, which we want to avoid.

I don't see any harm in having placeholder values for log potentials in logical factors. Currently it's None. But we can use any placeholder value. Once we have that we can actually all the factors the same way. In particular we can keep the previous implementation of register factor group, and no special processing is needed here.

pgmax/fg/graph.py
```diff
@@ -432,12 +526,23 @@ def __getitem__(self, name: Any) -> np.ndarray:
...
+                    f"The factor group {name} does not have log potentials"
+                )
+
             start = self.fg_state.factor_group_to_potentials_starts[factor_group]
             log_potentials = value[
                 start : start + factor_group.factor_group_log_potentials.shape[0]
             ]
         elif frozenset(name) in self.fg_state.variables_to_factors:
             factor = self.fg_state.variables_to_factors[frozenset(name)]
+            if not isinstance(factor, enumeration.EnumerationFactor):
```
StannisZhou, line 543: Same as above. Preferably no special processing.

pgmax/fg/graph.py
```diff
@@ -806,10 +910,23 @@ def BP(
         \tget_bp_state: Function to reconstruct the BPState from BPArrays.\n
         \tget_beliefs: Function to calculate beliefs from BPArrays.\n
     """
-    max_msg_size = int(np.max(bp_state.fg_state.wiring.edges_num_states))
-    num_val_configs = (
-        int(bp_state.fg_state.wiring.factor_configs_edge_states[-1, 0]) + 1
-    )
+    wiring = bp_state.fg_state.wiring
+    max_msg_size = int(np.max(wiring.edges_num_states))
```
StannisZhou, line 914: Calculate this on demand instead of having to define an additional `GraphWiring` class.

pgmax/fg/graph.py
```diff
@@ -806,10 +910,23 @@ def BP(
...
+
+    num_val_configs = 0
+    if "EnumerationFactor" in wiring.wiring_by_factor_subtype:
```
StannisZhou, line 917: No special processing. Add more abstractions and rely on generic structures instead.

In this case we need to pre-compute `num_val_configs` because it's a static argument so we need its concrete value. For some factors we need this type of concrete arguments. For others we don't. This is an abstraction we should make: for each factor type we should have a function that return the necessary concrete arguments.

One convenient place to put this is as a property `static_arguments` of the corresponding wiring class, which in this case would be a dictionary `{'num_val_configs': self.factor_configs_edge_states[-1, 0]}`.

And in this section what we would do is
```
static_arguments = {
	factor_type: wiring[factor_type].static_arguments
	for factor_type in wiring
}
```

pgmax/fg/graph.py
```diff
@@ -861,14 +977,28 @@ def update(msgs: jnp.ndarray, _) -> Tuple[jnp.ndarray, None]:
                 evidence,
                 wiring.var_states_for_edges,
             )
+            ftov_msgs = vtof_msgs
```
StannisZhou, line 980: If this is for initialization, `ftov_msgs = jnp.zeros_like(vtof_msgs)` might be better and less confusing.

pgmax/fg/graph.py
```diff
@@ -861,14 +977,28 @@ def update(msgs: jnp.ndarray, _) -> Tuple[jnp.ndarray, None]:
...
+                        num_val_configs,
+                        temperature,
+                    )
+                elif isinstance(factor_wiring, logical.LogicalWiring):
+                    ftov_msgs = logical.INFERENCE_BY_SUBTYPES[factor_subtype](
+                        ftov_msgs,
+                        factor_wiring.parents_edge_states,
+                        factor_wiring.children_edge_states,
+                        temperature,
+                    )
```
StannisZhou, line 1001: No special processing.

We should have a centralized dictionary, mapping factor types to their corresponding factor to variable updates.

For the function arguments, each factor to variable message updates function takes as input `ftov_msgs`, `log_potentials` and `temperature`. It additionally takes some static arguments (which we already have) and some non-static arguments.

For each wiring, we can have another property `nonstatic_arguments`. For an enumeration wiring it would be just `{'factor_configs_edge_states': self.factor_configs_edge_states}`.

Once we define these, at the beginning of the function we can have
```
arguments = {
	factor_type: {
		'log_potentials': log_potentials,
		'temperature': temperature,
		**static_arguments[factor_type], 
		**wiring[factor_type].nonstatic_arguments,
	}
	for factor_type in wiring
}
```
and you do the updates by
```
for factor_type in FAC_TO_VAR_UPDATES:
	ftov_msgs = FAC_TO_VAR_UPDATES[factor_type](**{'ftov_msgs': ftov_msgs, **arguments[factor_type]})
```
where `FAC_TO_VAR_UPDATES` is defined in some centralized place.

StannisZhou (commented):
Left more comments. The current version is unfortunately still quite problematic. Probably need a few more iterations.....

Most major issues are in `graph.py`. There are still many specialized processing, and it is simply not true that when we add a new factor we do not need to touch this file.

I left some fairly concrete comments on how we can make this work. To summarize here (and to add a few unmentioned points):
1. Strings don't seem like a good way of referring to the factors. It looks like we can just use factor types which might be better. JAX is doing that for registering pytrees internally https://github.com/google/jax/blob/e1fd6304d879142872cb2b5e761f2d206287919f/jax/_src/tree_util.py#L212
2. Almost all functions of `FactorGraph` should return a dictionary mapping factor types to corresponding values. I included a few examples.
3. We can adopt some placeholder values for logical factors log potentials. It doesn't hurt anything and by doing it we no longer need to worry about any special processing. In particular, you can revert the changes on various log potentials related values and functions (e.g. `_register_factor_group`). And be sure to fix the issue on the `factor_groups` property and get rid of `eq=False`.
4. We need to make more abstractions so that everything can be done by simple for loops. I made some detailed comments.
5. Please further consolidate things related to different factors into the factor module, including the factor to variable message update functions and the wiring concatenation functions. We can have a centralized registration in existing modules (e.g. in `bp/infer.py` we can have a `FAC_TO_VAR_UPDATES` dictionary mapping factor types to message update functions).

pgmax/fg/graph.py
```diff
@@ -160,51 +157,79 @@ def _register_factor_group(
...
-            self._factor_to_potentials_starts[factor] = (
-                self._factor_group_to_potentials_starts[factor_group]
-                + vv * factor.log_potentials.shape[0]
-            )
-            factor_group_num_configs += factor.log_potentials.shape[0]
 
-        self._total_factor_num_states += factor_num_states_cumsum[-1]
-        self._total_factor_num_configs += factor_group_num_configs
+            factor_subtype = factor.__subtype__
+            if factor_subtype not in self._subtypes_to_factors:
```
StannisZhou, line 170: Instead of this if, we should have a centralized list of all registered factor types right now, and when initializing the factor graph, in `__post_init__` we should already initialize `self._factor_type_to_factors = {factor_type: collections.OrderedDict()}`.

The idea is that all returned dictionaries would have one entry per factor type, regardless of whether there is a factor of that type in the factor graph. It just makes things uniform and reduces special processing in the form of this kind of ifs.

pgmax/factors/logical.py
```diff
@@ -0,0 +1,125 @@
...
+            The child variable's state 1 is children_edge_states[ii, 1] + 1.
+    """
+
+    parents_edge_states: Union[np.ndarray, jnp.ndarray]
+    children_edge_states: Union[np.ndarray, jnp.ndarray]
+
+
+@dataclass(frozen=True, eq=False)
+class LogicalFactor(nodes.Factor):
+    """A logical OR/AND factor
```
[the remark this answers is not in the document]
antoine-dedieu: I want to write a proper latex with proper derivations or the or / and with any temperature. For now I'll point to the HCN supplementary

pgmax/fg/graph.py
```diff
@@ -76,56 +87,42 @@ def __post_init__(self):
...
         return hash(self.factor_groups)
 
-    def add_factor(
-        self,
-        variable_names: List,
-        factor_configs: np.ndarray,
-        log_potentials: Optional[np.ndarray] = None,
-        name: Optional[str] = None,
-    ) -> None:
+    def add_factor(self, variable_names: List, factor_type: str, **kwargs) -> None:
```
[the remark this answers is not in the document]
antoine-dedieu: Why don't we set `factor_type` to `EnumerationFactor` as a default argument

pgmax/fg/fg_utils.py
```diff
@@ -1,50 +1,99 @@
...
 import numpy as np
 
-from pgmax.fg.nodes import EnumerationWiring
+from pgmax.factors.enumeration import EnumerationWiring
+from pgmax.factors.logical import LogicalWiring
 
 
 def concatenate_enumeration_wirings(
-    wirings: Sequence[EnumerationWiring],
+    enum_wirings: Sequence[EnumerationWiring], enum_factor_to_msgs_starts: np.ndarray
```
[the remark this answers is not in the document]
antoine-dedieu: We should discuss this. See below

pgmax/fg/graph.py
antoine-dedieu: This is necessary because `wiring` is computing `_factor_to_msgs_starts`

pgmax/fg/graph.py
```diff
@@ -160,51 +157,79 @@ def _register_factor_group(
...
+            concatenated_wiring_by_factor_subtype = CONCATENATE_WIRING[factor_subtype](
+                wirings_by_subtype, factor_to_msgs_starts_by_subtype
+            )
+            wiring_by_factor_subtype[
+                factor_subtype
+            ] = concatenated_wiring_by_factor_subtype
+
+        self._total_factor_num_states = factor_num_states_cumsum
+
+        graph_wiring = GraphWiring(
```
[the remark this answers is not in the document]
antoine-dedieu: I am fine with doing some thing like this. However `CONCATENATE_WIRINGS` needs to compute the `factor_to_msgs_starts`. Previously this computation was done independently in two separate parts, which is not what we want

pgmax/fg/graph.py
```diff
@@ -216,15 +241,37 @@ def log_potentials(self) -> np.ndarray:
...
-            ]
-        )
+        # Aggregate log potentials
+        factor_group_num_configs = 0
+        self._factor_group_to_potentials_starts = collections.OrderedDict()
+        self._factor_to_potentials_starts = collections.OrderedDict()
+
+        log_potentials = []
+        # Only EnumerationFactors have log potentials
+        if "EnumerationFactor" in self._subtypes_to_factors:
```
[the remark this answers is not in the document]
antoine-dedieu: I disagree with this one. We don't want to concatenate `None`. Maybe we should return a dict of log_potentials by factor_type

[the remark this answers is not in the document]
StannisZhou: This is now changed to concatenate empty numpy arrays

pgmax/fg/graph.py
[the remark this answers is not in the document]
StannisZhou: I changed the function name to `add_factor_by_type`. It's clearer for the user (as it keeps the current detailed docstring) and we also don't break any of the existing materials.

StannisZhou (commented):
Mostly minor comments. We can merge this after these comments are addressed!

Before merging, could you also check to make sure all the current examples (e.g. RBM, RCN, GMRF) still run fine? In the unit test only the ising model example is checked.

antoine-dedieu:
@StannisZhou with the recent changes:
- all the old + new unit tests pass
- coverage is at 100%
- the RCN, RBM, Ising, GMRF examples run fine

pgmax/fg/graph.py
```diff
@@ -129,12 +129,48 @@ def add_factor(
...
+    def add_factor_by_type(
+        self, variable_names: List, factor_type: type, *args, **kwargs
+    ) -> None:
+        """Function to add a single factor to the FactorGraph.
+
+        Args:
+            variable_names: A list containing the connected variable names.
+                Variable names are tuples of the type (variable_group_name, variable_name_within_variable_group)
+            factor_type: Type of factor to be added
+            args: Args to be passed to the factory function.
```
StannisZhou, line 141:
```suggestion
            args: Args to be passed to factor_type.
```

pgmax/fg/graph.py
```diff
@@ -129,12 +129,48 @@ def add_factor(
...
+            kwargs: kwargs to be passed to the factory function, and an optional "name" argument
+                for specifying the name of a named factor group.
```
StannisZhou, line 143:
```suggestion
            kwargs: kwargs to be passed to the factor_type, and an optional "name" argument
                for specifying the name of a named factor.
```

pgmax/fg/graph.py
```diff
@@ -129,12 +129,48 @@ def add_factor(
...
+        )
+        self._register_factor_group(factor_group, name)
+
     def add_factor_group(self, factory: Callable, *args, **kwargs) -> None:
         """Add a factor group to the factor graph
 
         Args:
             factory: Factory function that takes args and kwargs as input and outputs a factor group.
-            args: Args to be passed to the factory function.
+            args: Args to be passed to the factor type function.
```
StannisZhou, line 173:
```suggestion
            args: Args to be passed to the factory function.
```

StannisZhou (approved):
LGTM after fixing the docstrings. Excited to have this merged!

PR #129, opened 2022-03-29
Speeding up adding factors + computing wiring

antoine-dedieu:
Main goal of this PR is to speed up the process of adding factors to a FactorGraph and to compile wiring. In particular
- we do not instantiate factors in a FactorGraph by default. Instead all the computations, including wiring, happen at the factor groups level. 
- all the factors-related operations are computed on demand, when the user requires them
- we create a subfolder `pgmax/groups/` to clarify the structure

Before:
- adding factors for the RBM exp takes 25s, building run_bp takes 45s
- adding factors for the convor exp takes 20s, building run_bp takes 25s

After:
- adding factors for the RBM exp takes 2s, building run_bp takes 4s
- adding factors for the convor exp takes 2s, building run_bp takes 10s

Our next step will accelerate the wiring compilation further by  using `numba`


pgmax/fg/graph.py
```diff
@@ -31,10 +30,19 @@
 from jax.scipy.special import logsumexp
 
 from pgmax.bp import infer
-from pgmax.factors import FAC_TO_VAR_UPDATES
+from pgmax.factors import FAC_TO_VAR_UPDATES, enumeration, logical
 from pgmax.fg import groups, nodes
 from pgmax.utils import cached_property
 
+GROUPS_TO_TYPES: OrderedDict[Type, Type] = collections.OrderedDict(
```
StannisZhou, line 37: Might be better to make this an attribute of the corresponding factor group classes instead of having to maintain a centralized registry, something like
```
factor_type: type = field(init=False, default=enumeration.EnumerationFactor)
```

pgmax/fg/graph.py
```diff
@@ -146,7 +151,7 @@ def add_factor_by_type(
             To add an ORFactor to a FactorGraph fg, run::
 
                 fg.add_factor_by_type(
-                    variable_names=variables_names_for_OR_factor,
+                    variable_names_for_factors=[variables_names_for_OR_factor],
```
StannisZhou, line 154: Why this change? This function is meant to add single factors imperatively, so there should only be one set of `variable_names`

StannisZhou: I see it's to make all factor groups have the same consistent interface. That is fine, but to avoid confusion, the argument for `add_factor_by_type` should still be just `variable_names`. Inside the function we can wrap it inside a list and pass the list to `SingleFactorGroup`.

After the change `SingleFactorGroup` should have an additional check to make sure its `len(variable_names_for_factors) == 1`.

pgmax/fg/groups.py
```diff
@@ -556,21 +554,34 @@ class FactorGroup:
...
 
-    def __post_init__(self) -> None:
-        """Initializes a tuple of all the factors contained within this FactorGroup."""
-        object.__setattr__(
-            self,
-            "_variables_to_factors",
-            MappingProxyType(self._get_variables_to_factors()),
-        )
+    def __post_init__(self):
+        object.__setattr__(self, "num_factors", len(self.variable_names_for_factors))
```
StannisZhou, line 566: If you do this you might want to add `num_factors: int = field(init=False)` to the class definition. Otherwise mypy will complain when you re-enable the check

pgmax/fg/groups.py
```diff
@@ -556,21 +554,34 @@ class FactorGroup:
...
+                elif isinstance(self.variable_group, CompositeVariableGroup):
+                    variable = self.variable_group.variable_group_container[
+                        variable_name[0]
+                    ][variable_name[1]]
+                num_states = variable.num_states
+                edges_num_states.append(num_states)
+                variables_and_num_states.append((variable, num_states))
+
+        object.__setattr__(self, "factor_edges_num_states", np.array(edges_num_states))
+        object.__setattr__(self, "variables_and_num_states", variables_and_num_states)
```
StannisZhou, line 584: Same as above. Add to class definition

pgmax/fg/groups.py
```diff
@@ -586,27 +597,52 @@ def __getitem__(self, variables: Union[Sequence, Collection]) -> nodes.Factor:
             ValueError: if the queried factor is not present in the factor group
         """
         variables = frozenset(variables)
-        if variables not in self._variables_to_factors:
+        if variables not in self._variables_to_factors():
```
StannisZhou, line 600: This is a property so you shouldn't have the `()` in the end

pgmax/fg/groups.py
```diff
@@ -586,27 +597,52 @@ def __getitem__(self, variables: Union[Sequence, Collection]) -> nodes.Factor:
...
+        """
+        return self._get_variables_to_factors()
+
     @cached_property
     def factor_group_log_potentials(self) -> np.ndarray:
+        """Flattened array of log potentials"""
+        return self.factor_group_log_potentials_full.flatten()
+
+    @cached_property
+    def factor_group_log_potentials_full(self) -> np.ndarray:
```
StannisZhou, line 629: Why is there a full version? If this is the non-flat version it should just be used to define the class, and shouldn't be a property

pgmax/fg/groups.py
```diff
@@ -653,13 +684,19 @@ class SingleFactorGroup(FactorGroup):
...
         factor: the single factor in the SingleFactorGroup
     """
 
-    variable_names: List
     factor: nodes.Factor
 
+    def __post_init__(self):
+        super().__post_init__()
+
+        if not len(self.variable_names_for_factors) == 0:
```
StannisZhou, line 695: Do you mean `== 1`?

pgmax/fg/groups.py
```diff
@@ -717,41 +753,59 @@ class EnumerationFactorGroup(FactorGroup):
...
+    def factor_group_log_potentials_full(self) -> np.ndarray:
+        """Function to compile potential array for the factor group
+
+        Returns:
+            An array of log potentials for the factor groups.
+
+        Raises:
+            ValueError: if the specified log_potentials is not of the right shape
+        """
+        factor_group_log_potentials_full = np.array(
```
StannisZhou, line 792: Isn't this exactly the same as `self.log_potentials`?

pgmax/fg/groups.py
```diff
@@ -761,14 +815,35 @@ def _get_variables_to_factors(
...
+
+        Args:
+            vars_to_starts: A dictionary that maps variables to their global starting indices
+                For an n-state variable, a global start index of m means the global indices
+                of its n variable states are m, m + 1, ..., m + n - 1
+
+        Returns:
+             EnumerationWiring for the EnumerationFactorGroup
+        """
+        return _compile_factor_group_wiring(
```
StannisZhou, line 839: Can we keep all the relevant logic inside this function, instead of defining a function elsewhere?

pgmax/fg/groups.py
```diff
@@ -921,33 +987,74 @@ def _get_variables_to_factors(
...
+                : self.log_potential_matrix.shape[-1],
             ]
             .transpose((1, 2, 0))
             .reshape((-1, 2))
         )
-        object.__setattr__(self, "log_potential_matrix", log_potential_matrix)
+        object.__setattr__(self, "factor_configs", factor_configs)
+
+    @cached_property
+    def factor_group_log_potentials_full(self) -> np.ndarray:
```
StannisZhou, line 1026: Again not sure this property is necessary

pgmax/fg/graph.py
antoine-dedieu: totally agree

pgmax/fg/nodes.py
```diff
@@ -1,5 +1,3 @@
-from __future__ import annotations
```
wlehrach, line 1: I wouldn't drop these (actually, I'd always add them) - there are all sorts of annoyances that can crop up with ordering of modules/functions inside files and type annotations.

StannisZhou: If mypy is not complaining then probably fine to get rid of some of these? It's only useful for some of the modules and with mypy check enforced we don't need to worry about missing failure cases

pgmax/groups/logical.py
```diff
@@ -0,0 +1,169 @@
...
+
+    Args:
+        edge_states_offset: Offset to go from a variable's relevant state to its other state
+            For ORFactors the edge_states_offset is 1.
+    """
+
+    edge_states_offset: int = field(init=False, default=1)
+    factor_type: Type = field(init=False, default=logical.ORFactor)
+
+    def _get_variables_to_factors(
```
StannisZhou, line 110: Since `factor_type` is an attribute, it looks like this function can be implemented just one within the `LogicalFactorGroup`, with factor creations done by `self.factor_type(...)`?

pgmax/groups/logical.py
```diff
@@ -0,0 +1,169 @@
...
+            this_var_states_for_edges = np.arange(
+                vars_to_starts[variable], vars_to_starts[variable] + num_states
+            )
+            var_states_for_edges.append(this_var_states_for_edges)
+        print(time.time() - start)
+
+        edges_num_states_cumsum = 0
+        parents_edge_states = []
+        children_edge_states = []
+        for factor_idx, variable_names_for_factor in enumerate(
```
StannisZhou, line 63: This seems really the same as the previous concatenation approach. The only difference is whether we create the individual factor objects. Curious to know what the performance difference is....

Also, since we are just manipulating arrays, it seems easy to make this whole function parallel using numba. Can be done in the next PR!

pgmax/groups/enumeration.py
```diff
@@ -0,0 +1,459 @@
...
+    All factors in the group are assumed to have the same set of valid configurations and
+    the same potential function. Note that the log potential function is assumed to be
+    uniform 0 unless the inheriting class includes a log_potentials argument.
+
+    Args:
+        factor_configs: Array of shape (num_val_configs, num_variables)
+            An array containing explicit enumeration of all valid configurations
+    """
+
+    factor_configs: np.ndarray = field(init=True)
```
StannisZhou, line 28: I'm guessing just `factor_configs: np.ndarray` is fine (since we are overriding parent class)?

pgmax/groups/enumeration.py
```diff
@@ -0,0 +1,459 @@
...
+    log_potentials: np.ndarray = field(init=True, default=np.empty((0,)))
```
StannisZhou, line 29: Get rid of `field`. Also why do we need to change the original `Optional[np.ndarray] = None` here? `None` makes more sense than empty in this case

pgmax/groups/enumeration.py
```diff
@@ -0,0 +1,459 @@
...
+
+        Args:
+            vars_to_starts: A dictionary that maps variables to their global starting indices
+                For an n-state variable, a global start index of m means the global indices
+                of its n variable states are m, m + 1, ..., m + n - 1
+
+        Returns:
+             EnumerationWiring for the EnumerationFactorGroup
+        """
+        return _compile_factor_group_wiring(
```
StannisZhou, line 94: This compiling function is really the same as the `compile_wiring` function in individual enumeration factors. There should not be two different implementations of the same function. Since we are trying to consolidate, can we define a function in `factors.enumeration` that handles the wiring compilation for all 3 cases?

StannisZhou: In addition, it still feels wrong to have to repeat this function multiple times.

One thought for possibly defining this function only once (inside the parent class): we still take the compile then concatenate approach, but we create a functional implementation for wiring compilation for individual factors (which can be used without instantiating individual factors). We then wrap things in numba to do the individual compilation in parallel before the concatenation.

The same approach should also work for LogicalFactors (and in factor any new factors we add) without change.

We can discuss more and try to tackle this in the next PR. For this PR fine to just consolidate all 3 places into one function.

pgmax/fg/groups.py
```diff
@@ -556,23 +353,48 @@ class FactorGroup:
...
 
     variable_group: Union[CompositeVariableGroup, VariableGroup]
-    _variables_to_factors: Mapping[FrozenSet, nodes.Factor] = field(init=False)
+    variable_names_for_factors: Sequence[List]
+    num_factors: int = field(init=False)
+    factor_edges_num_states: np.ndarray = field(init=False)
+    variables_and_num_states: OrderedDict[FrozenSet, nodes.Factor] = field(init=False)
+    factor_type: Type = field(init=False)
+    factor_configs: np.ndarray = field(init=False, default=None)
+    log_potentials: np.ndarray = field(init=False)
```
StannisZhou, line 372: Add `default=np.empty((0,))` and get rid of the `if` in `__post_init__`.

pgmax/fg/groups.py
```diff
@@ -556,23 +353,48 @@ class FactorGroup:
...
+            If none, it is assumed the log potential is uniform 0 and such an array is automatically
+            initialized.
     """
 
     variable_group: Union[CompositeVariableGroup, VariableGroup]
-    _variables_to_factors: Mapping[FrozenSet, nodes.Factor] = field(init=False)
+    variable_names_for_factors: Sequence[List]
+    num_factors: int = field(init=False)
+    factor_edges_num_states: np.ndarray = field(init=False)
+    variables_and_num_states: OrderedDict[FrozenSet, nodes.Factor] = field(init=False)
```
StannisZhou, line 369: It seems it's better to change these to `cached_property`s (e.g. `factor_edges_num_states` to be consistent with how it's done for individual factors, and `variables_for_factors` which is the parallel of `variables` in individual factors.

pgmax/groups/enumeration.py
```diff
@@ -0,0 +1,459 @@
...
+class EnumerationFactorGroup(groups.FactorGroup):
+    """Class to represent a group of EnumerationFactors.
+
+    All factors in the group are assumed to have the same set of valid configurations and
+    the same potential function. Note that the log potential function is assumed to be
+    uniform 0 unless the inheriting class includes a log_potentials argument.
+
+    Args:
+        factor_configs: Array of shape (num_val_configs, num_variables)
+            An array containing explicit enumeration of all valid configurations
```
StannisZhou, line 25: log_potentials should also be documented (move the documentation currently in the parent class to here)

pgmax/groups/enumeration.py
```diff
@@ -0,0 +1,459 @@
...
+        Raises:
+            ValueError if:
+                (1) The specified log_potential_matrix is not a 2D or 3D array.
+                (2) Some pairwise factors connect to less or more than 2 variables.
+                (3) The specified log_potential_matrix does not match the number of factors.
+                (4) The specified log_potential_matrix does not match the number of variable states of the
+                    variables in the factors.
+    """
+
+    factor_configs: np.array = field(init=False)
```
StannisZhou, line 203: If we get rid of `factor_configs` in the parent class we don't need to do this awkward thing here

pgmax/fg/graph.py
```diff
@@ -70,6 +70,17 @@ def __post_init__(self):
             0,
             0,
         )
+
+        # Useful objects to build the FactorGraph
+        self._factor_types_to_groups: OrderedDict[
+            Type, List[groups.FactorGroup]
+        ] = collections.OrderedDict(
+            [(factor_type, []) for factor_type in FAC_TO_VAR_UPDATES]
+        )
```
StannisZhou, line 79: Move this back to avoid unnecessary diffs

pgmax/fg/graph.py
```diff
@@ -79,30 +90,23 @@ def __post_init__(self):
...
             [
                 factor_group
-                for factor_groups_per_type in self.factor_groups.values()
+                for factor_groups_per_type in self._factor_types_to_groups.values()
                 for factor_group in factor_groups_per_type
             ]
         )
         return hash(all_factor_groups)
 
+    # TODO: should we merge add_factor and add_factor_by_type?
```
StannisZhou, line 104: Remove before merging

pgmax/fg/graph.py
```diff
@@ -79,30 +90,23 @@ def __post_init__(self):
...
     def add_factor(
         self,
         variable_names: List,
         factor_configs: np.ndarray,
-        log_potentials: Optional[np.ndarray] = None,
+        log_potentials: Optional[np.ndarray] = np.empty((0,)),
```
StannisZhou, line 109: Why change `None` to empty?

pgmax/fg/graph.py
```diff
@@ -196,23 +200,35 @@ def _register_factor_group(
                 f"A factor group with the name {name} already exists. Please choose a different name!"
             )
 
-        for variables, factor in factor_group._variables_to_factors.items():
-            if variables in self._variables_to_factors:
+        # TODO: do we want to remove this for speed?
```
StannisZhou, line 203: We probably don't want to remove. I added this check after accidentially adding the same factor twice and getting really confused about the results. I think it's a useful check....

pgmax/fg/graph.py
```diff
@@ -196,23 +200,35 @@ def _register_factor_group(
...
+                == self._factors_var_names_to_type[tuple_var_names_for_factor]
+            ):
                 raise ValueError(
-                    f"A factor involving variables {variables} already exists. Please merge the corresponding factors."
+                    f"A {factor_type} involving variables {tuple_var_names_for_factor} already exists. Please merge the corresponding factors."
                 )
-            self._variables_to_factors[variables] = factor
+            self._factors_var_names_to_type[tuple_var_names_for_factor] = factor_type
+
+        factor_group_type = factor_group.factor_type
```
StannisZhou, line 217: Nit: this is really just factor_type. Factor groups don't have types....

StannisZhou (commented):
Made some detailed comments, although nothing major.

It looks like there should be a way to use numba parallel and significantly speed up wiring compilation while unifying the compile wiring interface for all factor groups. So although the current factor group wiring compilation is still a bit annoying, we can go with what we have now after some additional cleanups (since we are likely going to rewrite many of these in the next PR).

pgmax/fg/groups.py
```diff
@@ -556,23 +353,48 @@ class FactorGroup:
...
+    factor_type: Type = field(init=False)
+    factor_configs: np.ndarray = field(init=False, default=None)
```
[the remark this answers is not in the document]
antoine-dedieu: This is for mypy.
`factor_configs` has to be specified for EnumerationFactor, so if we only keep the log_potentials in the parent Factor class, we will get `Non default argument cannot follow a default one`

pgmax/fg/groups.py
antoine-dedieu: I would advocate to not moving to `cached_property` to only loop through the factors once, and save time

pgmax/fg/nodes.py
```diff
@@ -87,31 +85,31 @@ def edges_num_states(self) -> np.ndarray:
         )
         return edges_num_states
 
-    def compile_wiring(self, vars_to_starts: Mapping[Variable, int]) -> Wiring:
-        """Compile wiring for the factor
+    @staticmethod
+    def concatenate_wirings(wirings: Sequence) -> Wiring:
```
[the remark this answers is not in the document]
antoine-dedieu: yes: both `logical.py` and `enumeration.py` have `concatenate_wirings` before `compile_wiring` so we should make this consistent in the parent class

pgmax/groups/enumeration.py
antoine-dedieu: the parent class will assing log_potentials to `np.empty((0,))` by default, so adding the None option is adding noise

pgmax/groups/enumeration.py
antoine-dedieu: Same, it's mypy related

pgmax/groups/logical.py
antoine-dedieu: This is much faster as we do not instantiate the factors objects.
Totally agree, the next PR will speed this up using numba!

tests/fg/test_graph.py
```diff
@@ -32,7 +35,7 @@ def test_factor_graph():
     with pytest.raises(
         ValueError,
         match=re.escape(
-            f"A factor involving variables {frozenset([0])} already exists."
+            f"A {enumeration_factor.EnumerationFactor} involving variables {tuple([0])} already exists."
```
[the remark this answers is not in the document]
antoine-dedieu: this means we need to change `_register_factor_group`? let's discuss this

pgmax/fg/graph.py
antoine-dedieu: As I mentioned, for more consistency

pgmax/groups/enumeration.py
antoine-dedieu: I agree that this function should only be called once.
I am planning to move this to numba in the next PR, let's keep it as it is for now

pgmax/fg/graph.py
antoine-dedieu: The comment `# See FactorGraphState docstrings for documentation on the following fields` does not apply to this object

antoine-dedieu:
Thanks for your thorough review @StannisZhou 
There are a few points we should discuss related to mypy behaviour
I totally agree that the next PR should be focused on speeding the factor group wiring using numba!

tests/fg/test_graph.py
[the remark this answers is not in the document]
StannisZhou: I updated `fg.graph` for this part. Please update this test accordingly!

The fix right now is not entirely correct for LogicalFactors, but a proper fix seems unnecessarily complicated. I will create a issue to keep track of this but for now we can go with the not entirely correct solution using frozenset

StannisZhou (approved):
Left a few more minor comments. LGTM after addressing the comments and making all tests pass!

PR #133, opened 2022-04-05
Numba speedup for wiring + log potentials

antoine-dedieu:
This PR is the continuation of https://github.com/vicariousinc/PGMax/pull/129 and part of our efforts to speed up the adding of FactorGroups and the wiring compilation.

As https://github.com/vicariousinc/PGMax/pull/129 has moved most of the wiring computation to the FactorGroup level, we can now use `numba` for fast computation of these wirings

As a result:
 - adding factors for the RBM exp takes 3s, building run_bp takes 1s
 - adding factors for the convor exp takes 2s, building run_bp takes 1s

### DIFF
--- a/pgmax/factors/enumeration.py
+++ b/pgmax/factors/enumeration.py
@@ -6,6 +6,7 @@ from typing import Mapping, Sequence, Tuple, Union
 
 import jax
 import jax.numpy as jnp
+import numba as nb
 import numpy as np
 
 from pgmax.bp import bp_utils
@@ -201,35 +202,60 @@ def compile_enumeration_wiring(
     Returns:
         The EnumerationWiring
     """
-    var_states_for_edges = []
-    for variables_for_factor in variables_for_factors:
-        for variable in variables_for_factor:
-            num_states = variable.num_states
-            this_var_states_for_edges = np.arange(
-                vars_to_starts[variable], vars_to_starts[variable] + num_states
-            )
-            var_states_for_edges.append(this_var_states_for_edges)
+    var_states_for_edges = np.empty(shape=(2 * len(variables_for_factors),), dtype=int)
+    start_indices = np.arange(len(variables_for_factors))
+    var_states = np.array(
+        [vars_to_starts[variable] for variable in variables_for_factors]
+    )
+    _compile_var_states_numba(var_states_for_edges, start_indices, var_states)
 
-    # Note: edges_starts corresponds to the factor_to_msgs_start for the LogicalFactors
-    edges_starts = np.insert(factor_edges_num_states.cumsum(), 0, 0)[:-1].reshape(
-        -1, factor_configs.shape[1]
+    num_configs, num_variables = factor_configs.shape
+    factor_configs_edge_states = np.empty(
+        (num_factors * num_configs * num_variables, 2), dtype=int
+    )
+    assert factor_edges_num_states.shape[0] == num_factors * num_variables
+    factor_edges_starts = np.insert(np.cumsum(factor_edges_num_states), 0, 0)
+    _compile_enumeration_wiring_numba(
+        factor_configs_edge_states, factor_configs, factor_edges_starts, num_factors
     )
 
-    factor_configs_edge_states = np.stack(
-        [
-            np.repeat(
-                np.arange(factor_configs.shape[0] * num_factors),
-                factor_configs.shape[1],
-            ),
-            (factor_configs[None] + edges_starts[:, None, :]).flatten(),
-        ],
-        axis=1,
-    )
     return EnumerationWiring(
         edges_num_states=factor_edges_num_states,
-        var_states_for_edges=np.concatenate(var_states_for_edges),
+        var_states_for_edges=var_states_for_edges,
         factor_configs_edge_states=factor_configs_edge_states,
     )
+
+
+@nb.jit(parallel=False, cache=True, fastmath=True, nopython=True)
+def _compile_var_states_numba(var_states_for_edges, start_indices, var_states):
+    for idx in start_indices:
+        var_states_for_edges[2 * idx] = var_states[idx]
+        var_states_for_edges[2 * idx + 1] = var_states[idx] + 1
+
+
+@nb.jit(parallel=False, cache=True, fastmath=True, nopython=True)
+def _compile_enumeration_wiring_numba(
+    factor_configs_edge_states, factor_configs, factor_edges_starts, num_factors
+):
+    num_configs, num_variables = factor_configs.shape
+
+    for factor_idx in nb.prange(num_factors):
+        for config_idx in range(num_configs):
+            factor_config_idx = num_configs * factor_idx + config_idx
+            factor_configs_edge_states[
+                num_variables
+                * factor_config_idx : num_variables
+                * (factor_config_idx + 1),
+                0,
+            ] = factor_config_idx
+
+            for var_idx in range(num_variables):
+                factor_configs_edge_states[
+                    num_variables * factor_config_idx + var_idx, 1
+                ] = (
+                    factor_edges_starts[num_variables * factor_idx + var_idx]
+                    + factor_configs[config_idx, var_idx]
+                )
 
 
 @functools.partial(jax.jit, static_argnames=("num_val_configs", "temperature"))
--- a/pgmax/factors/enumeration.py
+++ b/pgmax/factors/enumeration.py
@@ -230,7 +230,7 @@ def compile_enumeration_wiring(
 
 @nb.jit(parallel=False, cache=True, fastmath=True, nopython=True)
 def _compile_var_states_numba(var_states_for_edges, num_states_cumsum, var_states):
-    "Fast numba computation of the var_states_for_edges of an EnumerationWiring."
+    "Fast numba computation of the var_states_for_edges of a Wiring."
 
     for variable_idx in nb.prange(num_states_cumsum.shape[0] - 1):
         start_variable, end_variable = (
--- a/pgmax/factors/enumeration.py
+++ b/pgmax/factors/enumeration.py
@@ -46,28 +46,28 @@ class EnumerationFactor(nodes.Factor):
     """An enumeration factor
 
     Args:
-        configs: Array of shape (num_val_configs, num_variables)
+        factor_configs: Array of shape (num_val_configs, num_variables)
             An array containing an explicit enumeration of all valid configurations
         log_potentials: Array of shape (num_val_configs,)
             An array containing the log of the potential value for each valid configuration
 
     Raises:
         ValueError: If:
-            (1) The dtype of the configs array is not int
+            (1) The dtype of the factor_configs array is not int
             (2) The dtype of the potential array is not float
-            (3) Configs does not have the correct shape
+            (3) factor_configs does not have the correct shape
             (4) The potential array does not have the correct shape
-            (5) The configs array contains invalid values
+            (5) The factor_configs array contains invalid values
     """
 
-    configs: np.ndarray
+    factor_configs: np.ndarray
     log_potentials: np.ndarray
 
     def __post_init__(self):
-        self.configs.flags.writeable = False
-        if not np.issubdtype(self.configs.dtype, np.integer):
+        self.factor_configs.flags.writeable = False
+        if not np.issubdtype(self.factor_configs.dtype, np.integer):
             raise ValueError(
-                f"Configurations should be integers. Got {self.configs.dtype}."
+                f"Configurations should be integers. Got {self.factor_configs.dtype}."
             )
 
         if not np.issubdtype(self.log_potentials.dtype, np.floating):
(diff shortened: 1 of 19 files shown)


pgmax/factors/enumeration.py
```diff
@@ -179,12 +180,13 @@ def concatenate_wirings(wirings: Sequence[EnumerationWiring]) -> EnumerationWiri
 
 def compile_enumeration_wiring(
     factor_edges_num_states: np.ndarray,
-    variables_for_factors: Tuple[Tuple[nodes.Variable, ...], ...],
+    variables_for_factors: Tuple[nodes.Variable, ...],
```
StannisZhou, line 183: I previously changed this because using a tuple of tuples can help us get rid of several unnecessary attributes (e.g. `factor_sizes` which I got rid of, and `num_factors` which I forgot to get rid of. Unless making it a tuple of tuples somehow makes it slow, I think having as close to a minimal set of nonredundant arguments is better.

pgmax/fg/groups.py
```diff
@@ -359,18 +359,47 @@ class FactorGroup:
...
+        ValueError: if the FactorGroup does not contain a Factor
     """
 
     variable_group: Union[CompositeVariableGroup, VariableGroup]
     variable_names_for_factors: Sequence[List]
     factor_type: Type = field(init=False)
     factor_configs: np.ndarray = field(init=False)
     log_potentials: np.ndarray = field(init=False, default=np.empty((0,)))
+    factor_sizes: np.ndarray = field(init=False)
+    variables_for_factors: Tuple[nodes.Variable, ...] = field(init=False)
```
StannisZhou, line 380: I think making `variables_for_factors` a tuple of tuples (a parallel of `variable_names_for_factors`) is cleaner, and allows us to get rid of `factor_sizes`. Would probably also come in handy in anticipation to our variables-related refactors.

pgmax/fg/groups.py
```diff
@@ -359,18 +359,47 @@ class FactorGroup:
...
+        factor_edges_num_states: Array concatenating the number of states for the variables connected to each Factor of
+            the FactorGroup. Each variable will appear once for each Factor it connects to.
+
+    Raises:
+        ValueError: if the FactorGroup does not contain a Factor
     """
 
     variable_group: Union[CompositeVariableGroup, VariableGroup]
     variable_names_for_factors: Sequence[List]
     factor_type: Type = field(init=False)
```
StannisZhou, line 376: `factor_type` should also be an attribute instead of an argument, since it's never explicitly specified at instiatiation

StannisZhou (commented):
The current implementation is valid, but one issue is to write a new factor group the developer needs to understand various bits of wiring compilation. This seems unnecessary. Instead, it would be best if we implement the wiring compilation once within the parent `FactorGroup` class and everything is automatically taken care of in new factor groups.

One possible way to do this:
1. For each type of factor, implement a static method `compile_wiring_numba` which takes some inputs and do numba wiring compilation.
2. For each type of factor, specify a property `wiring_compilation_arguments`, similar to the current `inference_arguments` in the wirings.
3. In `FactorGroup`, add an optional `use_numba=True` flag to `compile_wiring`. If `use_numba` is `True`, we do something like:
```
wiring_compilation_arguments = {
	key: getattr(self, key) for key in self.factor_type.wiring_compilation_arguments
}
wiring = factor_type.compile_wiring_numba(
	vars_to_starts=vars_to_starts,
	**wiring_compilation_arguments
)
```
to get the wiring.
4. Get rid of all the customized wiring compilation implementations in the various factor groups.

In the process it would also be best if we can consolidate the number of arguments we need (for example if we make `variables_for_factors` as tuple of tuples we can get rid of `factor_sizes` and `num_factors`).

pgmax/factors/enumeration.py
antoine-dedieu: making it a tuple of tuple makes it slower, which is why I have changed it

pgmax/factors/enumeration.py
```diff
@@ -201,37 +203,72 @@ def compile_enumeration_wiring(
...
         edges_num_states=factor_edges_num_states,
-        var_states_for_edges=np.concatenate(var_states_for_edges),
+        var_states_for_edges=var_states_for_edges,
         factor_configs_edge_states=factor_configs_edge_states,
     )
 
 
+@nb.jit(parallel=False, cache=True, fastmath=True, nopython=True)
+def _compile_var_states_numba(var_states_for_edges, num_states_cumsum, var_states):
+    "Fast numba computation of the var_states_for_edges of a Wiring."
```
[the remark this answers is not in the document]
wlehrach: Also, can you add type hints

pgmax/groups/enumeration.py
```diff
@@ -429,3 +428,13 @@ def unflatten(
             )
 
         return data
+
+
+@nb.jit(parallel=False, cache=True, fastmath=True, nopython=True)
+def _compute_log_potentials(log_potentials, log_potential_matrix, factor_configs):
```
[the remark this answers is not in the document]
antoine-dedieu: there is no need to write docstring for internal functions
https://stackoverflow.com/questions/7687407/are-docstrings-for-internal-functions-python-necessary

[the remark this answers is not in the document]
antoine-dedieu: I'll mention the in-place mutations + will add type hints

antoine-dedieu:
@StannisZhou I am fine with the `wiring_compilation_arguments` but I do not think numba should be involved here.
A `factor_type.compile_wiring` may use an internal `_compile_wiring_numba` to speed things up, but that should not be specified at the `FactorGroup` level

StannisZhou:
Then we can change the existing compile_wiring function to be functional. Seems like with the wiring_compilation_arguments it wouldn't make things harder to use so I'm fine with that

pgmax/fg/groups.py
wlehrach: slower in general, or slower for numba?

antoine-dedieu: @wlehrach this is not related to numba but to the number of lists manipulated

Screen Shot 2022-04-06 at 11 45 16 AM

pgmax/factors/enumeration.py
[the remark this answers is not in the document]
StannisZhou: For type hints, does mypy play well with numba? I was reading online that they don't play well together so refrained from suggesting to add type hints

antoine-dedieu:
@StannisZhou I have pushed a commit that should address your comments. 
The two nice thing are that 
(1) we only define the `factor_group.compile_wiring` once, in the parent class
(2) we do not need to match the Factor and FactorGroup arguments, because wiring is always called at the FactorGroup level (even when it is a SingleFactorGroup, which was my concern)

However there is something weird with having the compile_wiring_arguments at the Factors level (note: I had to make this @staticmethod because a @property is not iterable)
because now 
(1) `factor.compile_wiring_arguments()` uses FactorGroup arguments
(2) we cannot call `factor.compile_wiring()`

pgmax/factors/logical.py
```diff
@@ -102,6 +104,19 @@ def __post_init__(self):
         if not np.all([variable.num_states == 2 for variable in self.variables]):
             raise ValueError("All variables should all be binary")
 
+    @staticmethod
+    def compile_wiring_arguments() -> List[str]:
```
StannisZhou, line 108: Remove and replace with `inspect.getargspec`

pgmax/factors/logical.py
```diff
@@ -146,26 +161,90 @@ def concatenate_wirings(wirings: Sequence[LogicalWiring]) -> LogicalWiring:
...
+
+        return LogicalWiring(
+            edges_num_states=factor_edges_num_states,
+            var_states_for_edges=var_states_for_edges,
+            parents_edge_states=parents_edge_states,
+            children_edge_states=children_edge_states,
+            edge_states_offset=edge_states_offset,
+        )
+
+    # def compile_wiring(
```
StannisZhou, line 227: Remember to remove commented code

pgmax/fg/nodes.py
```diff
@@ -85,6 +85,16 @@ def edges_num_states(self) -> np.ndarray:
         )
         return edges_num_states
 
+    @property
+    def compile_wiring_arguments(self) -> List[str]:
```
StannisZhou, line 89: Remove

pgmax/fg/nodes.py
```diff
@@ -99,17 +109,19 @@ def concatenate_wirings(wirings: Sequence) -> Wiring:
...
-                For an n-state variable, a global start index of m means the global indices
-                of its n variable states are m, m + 1, ..., m + n - 1
-
-        Returns:
-            Wiring for the Factor
-        """
-        raise NotImplementedError(
-            "Please subclass the Factor class and override this method"
-        )
+    # TODO: delete as this creates mypy issues
```
StannisZhou, line 112: What is the mypy issue?

pgmax/groups/enumeration.py
```diff
@@ -330,26 +308,26 @@ def _get_variables_to_factors(
...
-             EnumerationWiring for the PairwiseFactorGroup
-        """
-        return enumeration.compile_enumeration_wiring(
-            factor_edges_num_states=self.factor_edges_num_states,
-            variables_for_factors=self.variables_for_factors,
-            num_factors=self.num_factors,
-            factor_configs=self.factor_configs,
-            vars_to_starts=vars_to_starts,
-        )
+    # def compile_wiring(
```
StannisZhou, line 311: Remove commented code

StannisZhou (commented):
Main comment is to use `inspect` to get the necessary arguments. Other bits LGTM

pgmax/fg/nodes.py
antoine-dedieu: I get this
```
pgmax/factors/enumeration.py:160: error: Signature of "compile_wiring" incompatible with supertype "Factor"
pgmax/factors/enumeration.py:160: note:      Superclass:
pgmax/factors/enumeration.py:160: note:          @staticmethod
pgmax/factors/enumeration.py:160: note:          def compile_wiring(vars_to_starts: Mapping[Variable, int]) -> Wiring
pgmax/factors/enumeration.py:160: note:      Subclass:
pgmax/factors/enumeration.py:160: note:          @staticmethod
pgmax/factors/enumeration.py:160: note:          def compile_wiring(factor_edges_num_states: Any, variables_for_factors: Tuple[Variable, ...], factor_configs: Any, vars_to_starts: Mapping[Variable, int], num_factors: int) -> EnumerationWiring
pgmax/factors/logical.py:152: error: Signature of "compile_wiring" incompatible with supertype "Factor"
pgmax/factors/logical.py:152: note:      Superclass:
pgmax/factors/logical.py:152: note:          @staticmethod
pgmax/factors/logical.py:152: note:          def compile_wiring(vars_to_starts: Mapping[Variable, int]) -> Wiring
pgmax/factors/logical.py:152: note:      Subclass:
pgmax/factors/logical.py:152: note:          @staticmethod
pgmax/factors/logical.py:152: note:          def compile_wiring(factor_edges_num_states: Any, variables_for_factors: Tuple[Variable, ...], factor_sizes: Any, vars_to_starts: Mapping[Variable, int], edge_states_offset: int) -> LogicalWiring
Found 2 errors in 2 files (checked 31 source files)
```

pgmax/fg/groups.py
```diff
@@ -525,6 +526,14 @@ def __post_init__(self):
 
         object.__setattr__(self, "factor_type", type(self.factor))
 
+        if hasattr(self.factor, "configs"):
+            object.__setattr__(self, "factor_configs", self.factor.configs)
+
+        if hasattr(self.factor, "edge_states_offset"):
+            object.__setattr__(
+                self, "edge_states_offset", self.factor.edge_states_offset
+            )
```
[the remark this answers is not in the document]
antoine-dedieu: Because only EnumerationFactors have `configs` and only LogicalFactor have `edge_states_offset`.
However these arguments are needed for the SingleFactorGroup for inference

StannisZhou (approved):
One more minor comment. LGTM otherwise!

pgmax/fg/nodes.py
StannisZhou: It indeed seems tricky to solve, so I added a check in `__post_init__`.

pgmax/factors/enumeration.py
```diff
@@ -176,60 +152,115 @@ def concatenate_wirings(wirings: Sequence[EnumerationWiring]) -> EnumerationWiri
...
-        axis=1,
-    )
-    return EnumerationWiring(
-        edges_num_states=factor_edges_num_states,
-        var_states_for_edges=np.concatenate(var_states_for_edges),
-        factor_configs_edge_states=factor_configs_edge_states,
-    )
+
+@nb.jit(parallel=False, cache=True, fastmath=True, nopython=True)
+def _compile_var_states_numba(
```
wlehrach, line 215: Is there a reason you make the caller allocate these arrays?  In general t's cleaner and less likely to result in error to allocate return arrays inside numba rather that mutating a passed in array - just do it outside of the for loop.  You can get a very small optimization by re-using arrays between calls (so highly performance sensitive code it can be useful), but you're not doing that here.

Issue #124, opened 2022-03-03
Update `graph.BP` interface to be more intuitive


StannisZhou:
Update the interface to be more like [JAX optimizers](https://jax.readthedocs.io/en/latest/jax.example_libraries.optimizers.html), to allow for example repeatedly updating messages.

PR #135, opened 2022-04-07
Make BP closer to jax optimizer

antoine-dedieu:
Resolves https://github.com/vicariousinc/PGMax/issues/124

We make `graph.BP` closer to JAX optimizers https://jax.readthedocs.io/en/latest/jax.example_libraries.optimizers.html

### DIFF
--- a/pgmax/fg/graph.py
+++ b/pgmax/fg/graph.py
@@ -857,7 +857,7 @@ class BPArrays:
 class BPContainer(NamedTuple):
     """A container for belief propagation functions.
 
-    Attributes:
+    Arguments:
         init: Function for creating log_potentials, ftov_msgs and evidence.
         update: Function for updating log_potentials, ftov_msgs and evidence.
         run_bp: Function for running belief propagation for num_iters with a damping_factor.
--- a/pgmax/fg/graph.py
+++ b/pgmax/fg/graph.py
@@ -855,7 +855,7 @@ class BPArrays:
 
 @dataclass(frozen=True, eq=False)
 class BeliefPropagation:
-    """A container for belief propagation functions.
+    """Belief propagation functions.
 
     Arguments:
         init: Function to create log_potentials, ftov_msgs and evidence.
@@ -1082,14 +1082,14 @@ def BP(bp_state: BPState, temperature: float = 0.0) -> BeliefPropagation:
         )
         return beliefs
 
-    bp_container = BeliefPropagation(
+    bp = BeliefPropagation(
         init=functools.partial(update, None),
         update=update,
         run_bp=run_bp,
         to_bp_state=to_bp_state,
         get_beliefs=get_beliefs,
     )
-    return bp_container
+    return bp
 
 
 @jax.jit
--- a/pgmax/fg/graph.py
+++ b/pgmax/fg/graph.py
@@ -854,21 +854,20 @@ class BPArrays:
 
 
 def BP(
-    bp_state: BPState, num_iters: int, temperature: float = 0.0
-) -> Tuple[Callable, Callable, Callable]:
+    bp_state: BPState, temperature: float = 0.0
+) -> Tuple[Callable, Callable, Callable, Callable]:
     """Function for generating belief propagation functions.
 
     Args:
         bp_state: Belief propagation state.
-        num_iters: Number of belief propagation iterations.
         temperature: Temperature for loopy belief propagation.
             1.0 corresponds to sum-product, 0.0 corresponds to max-product.
 
     Returns:
         Tuple containing\n
-        \trun_bp: Function for running belief propagation for num_iters.\n
-        \t\tOptionally takes as input log_potentials updates, ftov_msgs updates,
-        \t\tevidence updates, and damping factor, and outputs a BPArrays.\n
+        \tbp_arrays_init: Function for updating log_potentials, ftov_msgs and tevidence
+        \t\tbefore running belief propagation.
+        \trun_bp: Function for running belief propagation for num_iters with a damping_factor.\n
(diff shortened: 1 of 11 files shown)


pgmax/fg/graph.py
```diff
@@ -1002,7 +1022,7 @@ def get_beliefs(bp_arrays: BPArrays) -> Any:
         )
         return beliefs
 
-    return run_bp, get_bp_state, get_beliefs
+    return bp_arrays_init, bp_arrays_update, get_bp_state, get_beliefs
```
StannisZhou, line 1025: Instead of returning a tuple, I suggest we bundle these into a dataclass (similar to https://github.com/deepmind/dm-haiku/blob/6f2492db3f06966adff4bb4ce0da3204c422c8b0/haiku/_src/transform.py#L40 but with dataclass not named tuple) with 4 callable attributes. Then the return type would make more sense/is easier to understand and the documentation can be done properly inside the new dataclass

pgmax/fg/graph.py
```diff
@@ -888,22 +887,18 @@ def BP(
     factor_type_to_msgs_range = bp_state.fg_state.factor_type_to_msgs_range
     factor_type_to_potentials_range = bp_state.fg_state.factor_type_to_potentials_range
 
-    def run_bp(
+    def bp_arrays_init(
```
StannisZhou, line 890: Instead of `init`, a better name here might actually be `update_bp_arrays` (since we are literally updating `bp_arrays`). And this function should take an additional optional argument `bp_arrays`. When the additional argument is `None` we grab values from `bp_state`.

pgmax/fg/graph.py
```diff
@@ -924,6 +919,31 @@ def run_bp(
         if evidence_updates is not None:
             evidence = update_evidence(evidence, evidence_updates, bp_state.fg_state)
 
+        return BPArrays(
+            log_potentials=log_potentials, ftov_msgs=ftov_msgs, evidence=evidence
+        )
+
+    def bp_arrays_update(
```
StannisZhou, line 926: Calling this `run_bp` might be more appropriate. This is not any kind of update, but only updating the messages by running BP

pgmax/fg/graph.py
StannisZhou: We can also change `get_bp_state` to `bp_arrays_to_bp_state` or something similar to make the function easier to understand

StannisZhou: In addition, here we should return a dataclass bundling 5 functions: the 4 existing ones (`update_bp_arrays`, `run_bp`, `bp_arrays_to_bp_state`, `get_beliefs`), and an additional `init_bp_arrays` (to be consistent having verb first is better) which can be defined as `functools.partial(update_bp_arrays, bp_arrays=None)`.

StannisZhou (commented):
Made a few comments. Main things to do are:
1. Bundle 5 functions into a dataclass for better organization/documentation and easier usage.
2. Do some renaming.

StannisZhou:
Also if you want the corresponding issue to be automatically linked you need to say resolves https://github.com/vicariousinc/PGMax/issues/124

antoine-dedieu:
@StannisZhou all your comments have been adressed

pgmax/fg/graph.py
```diff
@@ -853,24 +854,34 @@ def tree_unflatten(cls, aux_data, children):
         return cls(**aux_data.unflatten(children))
 
 
-def BP(
-    bp_state: BPState, num_iters: int, temperature: float = 0.0
-) -> Tuple[Callable, Callable, Callable]:
+class BPContainer(NamedTuple):
```
StannisZhou, line 857: Container sounds weird...How about just `BeliefPropagation`? Also shorten all the existing `bp_container` to just `bp`.

pgmax/fg/graph.py
```diff
@@ -853,24 +854,34 @@ def tree_unflatten(cls, aux_data, children):
...
+    """A container for belief propagation functions.
+
+    Arguments:
+        init: Function for creating log_potentials, ftov_msgs and evidence.
+        update: Function for updating log_potentials, ftov_msgs and evidence.
+        run_bp: Function for running belief propagation for num_iters with a damping_factor.
+        get_bp_state: Function to reconstruct the BPState from BPArrays.
+        get_beliefs: Function to calculate beliefs from BPArrays.
```
StannisZhou, line 865: Copy the respective docstrings here so that people know how to use these? The docstrings in individual functions won't be accessible in the documentation. Only the docstrings here would be

pgmax/fg/graph.py
StannisZhou: Our internal software criteria says use dataclasses not named tuples. Follow those and use dataclasses for now?

pgmax/fg/graph.py
```diff
@@ -1002,7 +1044,14 @@ def get_beliefs(bp_arrays: BPArrays) -> Any:
         )
         return beliefs
 
-    return run_bp, get_bp_state, get_beliefs
+    bp_container = BPContainer(
+        init=functools.partial(update, None),
```
StannisZhou, line 1048: Does this mean `init` can only take keyword arguments? If so mention that in the documentation for `init` in `BeliefPropagation`

StannisZhou (approved):
LGTM after addressing minor comments:
1. Change `BPContainer` to `BeliefPropagation` (or maybe `PropagationFunctions`?)
2. Shorten all `bp_container` to just `bp`
3. Use `dataclass` instead of `NamedTuple`
4. Update docstrings

pgmax/fg/graph.py
antoine-dedieu: this restriction does not apply, why do you think it would?

PR #136, opened 2022-04-13
Variables refactor

antoine-dedieu:
We update the way of representing variables. In particular:
- We get rid of variables names, as welll as of the `Variables` and `CompositeVariableGroup` classes. A variable is now represented by a tuple `(variable hash, variable num_states)`
  In particular, a FactorGraph can then directly be instantiated as`fg = graph.FactorGraph(variables=[hidden_variables, visible_variables])`
  Similarly,  Factors are defined by directly passing the variables involved, as `[hidden_variables[ii], visible_variables[jj]]`
- We rewrite `NDVariableArray` so that the user can access variables by relying on the use of numpy arrays. We also optimize some follow-up computations.

### DIFF
--- a/examples/ising_model.py
+++ b/examples/ising_model.py
@@ -29,7 +29,7 @@ from pgmax.groups import variables as vgroup
 
 # %%
 variables = vgroup.NDVariableArray(num_states=2, shape=(50, 50))
-fg = graph.FactorGraph(variables=variables)
+fg = graph.FactorGraph(variable_groups=variables)
 
 variables_for_factors = []
 for ii in range(50):
@@ -43,7 +43,7 @@ factor_group = enumeration.PairwiseFactorGroup(
     variables_for_factors=variables_for_factors,
     log_potential_matrix=0.8 * np.array([[1.0, -1.0], [-1.0, 1.0]]),
 )
-fg.add_factor_group(factor_group)
+fg.add_factors(factor_group)
 
 # %% [markdown]
 # ### Run inference and visualize results
--- a/examples/ising_model.py
+++ b/examples/ising_model.py
@@ -39,12 +39,11 @@ for ii in range(50):
         variables_for_factors.append([variables[ii, jj], variables[kk, jj]])
         variables_for_factors.append([variables[ii, jj], variables[kk, ll]])
 
-fg.add_factor_group(
-    factory=enumeration.PairwiseFactorGroup,
+factor_group = enumeration.PairwiseFactorGroup(
     variables_for_factors=variables_for_factors,
     log_potential_matrix=0.8 * np.array([[1.0, -1.0], [-1.0, 1.0]]),
-    name="factors",
 )
+fg.add_factor_group(factor_group, name="factors")
 
 # %% [markdown]
 # ### Run inference and visualize results
--- a/examples/ising_model.py
+++ b/examples/ising_model.py
@@ -46,25 +46,19 @@ fg.add_factor_group(
     log_potential_matrix=0.8 * np.array([[1.0, -1.0], [-1.0, 1.0]]),
 )
 
-# %%
-from pgmax.factors import enumeration as enumeration_factor
-
-factors = fg.factor_groups[enumeration_factor.EnumerationFactor][0].factors
-
 # %% [markdown]
 # ### Run inference and visualize results
 
-import imp
-
 # %%
-from pgmax.fg import graph
-
-imp.reload(graph)
 bp = graph.BP(fg.bp_state, temperature=0)
 
 # %%
+d = {variables: 1, variables.__hash__(): 2}
+hd = graph.HashableDict(d)
+hd[variables]
+
+# %%
 # TODO: check\ time for before BP vs time for BP
-# TODO: why bug when done twice?
 # TODO: time PGMAX vs PMP
 bp_arrays = bp.init(
     evidence_updates={variables: jax.device_put(np.random.gumbel(size=(50, 50, 2)))}
--- a/examples/ising_model.py
+++ b/examples/ising_model.py
@@ -43,7 +43,7 @@ factor_group = enumeration.PairwiseFactorGroup(
     variables_for_factors=variables_for_factors,
     log_potential_matrix=0.8 * np.array([[1.0, -1.0], [-1.0, 1.0]]),
 )
-fg.add_factor_group(factor_group, name="factors")
+fg.add_factor_group(factor_group)
 
 # %% [markdown]
 # ### Run inference and visualize results
@@ -88,7 +88,7 @@ batch_loss(None, {variables: jax.device_put(np.random.gumbel(size=(10, 50, 50, 2
 
 # %%
 grads = log_potentials_grads(
-    {"factors": jnp.eye(2)},
+    {factor_group: jnp.eye(2)},
     {variables: jax.device_put(np.random.gumbel(size=(50, 50, 2)))},
 )
 
--- a/examples/ising_model.py
+++ b/examples/ising_model.py
@@ -37,7 +37,7 @@ for ii in range(50):
         kk = (ii + 1) % 50
         ll = (jj + 1) % 50
         variables_for_factors.append([variables[ii, jj], variables[kk, jj]])
-        variables_for_factors.append([variables[ii, jj], variables[kk, ll]])
+        variables_for_factors.append([variables[ii, jj], variables[ii, ll]])
 
 factor_group = enumeration.PairwiseFactorGroup(
     variables_for_factors=variables_for_factors,
@@ -52,8 +52,6 @@ fg.add_factors(factor_group)
 bp = graph.BP(fg.bp_state, temperature=0)
 
 # %%
-# TODO: check\ time for before BP vs time for BP
-# TODO: time PGMAX vs PMP
 bp_arrays = bp.init(
     evidence_updates={variables: jax.device_put(np.random.gumbel(size=(50, 50, 2)))}
 )
--- a/examples/ising_model.py
+++ b/examples/ising_model.py
@@ -31,19 +31,19 @@ from pgmax.groups import variables as vgroup
 variables = vgroup.NDVariableArray(num_states=2, shape=(50, 50))
 fg = graph.FactorGraph(variables=variables)
 
-# TODO: rename variable_for_factors?
-variable_names_for_factors = []
+variables_for_factors = []
 for ii in range(50):
     for jj in range(50):
         kk = (ii + 1) % 50
         ll = (jj + 1) % 50
-        variable_names_for_factors.append([variables[ii, jj], variables[kk, jj]])
-        variable_names_for_factors.append([variables[ii, jj], variables[kk, ll]])
+        variables_for_factors.append([variables[ii, jj], variables[kk, jj]])
+        variables_for_factors.append([variables[ii, jj], variables[kk, ll]])
 
 fg.add_factor_group(
     factory=enumeration.PairwiseFactorGroup,
-    variable_names_for_factors=variable_names_for_factors,
+    variables_for_factors=variables_for_factors,
     log_potential_matrix=0.8 * np.array([[1.0, -1.0], [-1.0, 1.0]]),
+    name="factors",
 )
 
 # %% [markdown]
@@ -53,21 +53,16 @@ fg.add_factor_group(
 bp = graph.BP(fg.bp_state, temperature=0)
 
 # %%
-d = {variables: 1, variables.__hash__(): 2}
-hd = graph.HashableDict(d)
-hd[variables]
-
-# %%
 # TODO: check\ time for before BP vs time for BP
 # TODO: time PGMAX vs PMP
 bp_arrays = bp.init(
     evidence_updates={variables: jax.device_put(np.random.gumbel(size=(50, 50, 2)))}
 )
 bp_arrays = bp.run_bp(bp_arrays, num_iters=3000)
-output = bp.get_bp_output(bp_arrays)
+beliefs = bp.get_beliefs(bp_arrays)
 
 # %%
-img = output.map_states[variables]
+img = graph.decode_map_states(beliefs)[variables]
 fig, ax = plt.subplots(1, 1, figsize=(10, 10))
 ax.imshow(img)
 
@@ -82,19 +77,20 @@ def loss(log_potentials_updates, evidence_updates):
     )
     bp_arrays = bp.run_bp(bp_arrays, num_iters=3000)
     beliefs = bp.get_beliefs(bp_arrays)
-    loss = -jnp.sum(beliefs)
+    loss = -jnp.sum(beliefs[variables])
     return loss
 
 
-batch_loss = jax.jit(jax.vmap(loss, in_axes=(None, {None: 0}), out_axes=0))
+batch_loss = jax.jit(jax.vmap(loss, in_axes=(None, {variables: 0}), out_axes=0))
 log_potentials_grads = jax.jit(jax.grad(loss, argnums=0))
 
 # %%
-batch_loss(None, {None: jax.device_put(np.random.gumbel(size=(10, 50, 50, 2)))})
+batch_loss(None, {variables: jax.device_put(np.random.gumbel(size=(10, 50, 50, 2)))})
 
 # %%
 grads = log_potentials_grads(
-    {"factors": jnp.eye(2)}, {None: jax.device_put(np.random.gumbel(size=(50, 50, 2)))}
+    {"factors": jnp.eye(2)},
+    {variables: jax.device_put(np.random.gumbel(size=(50, 50, 2)))},
 )
 
 # %% [markdown]
@@ -104,15 +100,15 @@ grads = log_potentials_grads(
 bp_state = bp.to_bp_state(bp_arrays)
 
 # Query evidence for variable (0, 0)
-bp_state.evidence[0, 0]
+bp_state.evidence[variables[0, 0]]
 
 # %%
 # Set evidence for variable (0, 0)
-bp_state.evidence[0, 0] = np.array([1.0, 1.0])
-bp_state.evidence[0, 0]
+bp_state.evidence[variables[0, 0]] = np.array([1.0, 1.0])
+bp_state.evidence[variables[0, 0]]
 
 # %%
 # Set evidence for all variables using an array
 evidence = np.random.randn(50, 50, 2)
-bp_state.evidence[None] = evidence
-bp_state.evidence[10, 10] == evidence[10, 10]
+bp_state.evidence[variables] = evidence
+np.allclose(bp_state.evidence[variables[10, 10]], evidence[10, 10])
--- a/examples/ising_model.py
+++ b/examples/ising_model.py
@@ -30,35 +30,50 @@ from pgmax.groups import variables as vgroup
 # %%
 variables = vgroup.NDVariableArray(num_states=2, shape=(50, 50))
 fg = graph.FactorGraph(variables=variables)
+
+# TODO: rename variable_for_factors?
 variable_names_for_factors = []
 for ii in range(50):
     for jj in range(50):
         kk = (ii + 1) % 50
         ll = (jj + 1) % 50
-        variable_names_for_factors.append([(ii, jj), (kk, jj)])
-        variable_names_for_factors.append([(ii, jj), (ii, ll)])
+        variable_names_for_factors.append([variables[ii, jj], variables[kk, jj]])
+        variable_names_for_factors.append([variables[ii, jj], variables[kk, ll]])
 
 fg.add_factor_group(
     factory=enumeration.PairwiseFactorGroup,
     variable_names_for_factors=variable_names_for_factors,
     log_potential_matrix=0.8 * np.array([[1.0, -1.0], [-1.0, 1.0]]),
-    name="factors",
 )
+
+# %%
+from pgmax.factors import enumeration as enumeration_factor
+
+factors = fg.factor_groups[enumeration_factor.EnumerationFactor][0].factors
 
 # %% [markdown]
 # ### Run inference and visualize results
 
+import imp
+
 # %%
+from pgmax.fg import graph
+
+imp.reload(graph)
 bp = graph.BP(fg.bp_state, temperature=0)
 
 # %%
+# TODO: check\ time for before BP vs time for BP
+# TODO: why bug when done twice?
+# TODO: time PGMAX vs PMP
 bp_arrays = bp.init(
-    evidence_updates={None: jax.device_put(np.random.gumbel(size=(50, 50, 2)))}
+    evidence_updates={variables: jax.device_put(np.random.gumbel(size=(50, 50, 2)))}
 )
 bp_arrays = bp.run_bp(bp_arrays, num_iters=3000)
+output = bp.get_bp_output(bp_arrays)
 
 # %%
-img = graph.decode_map_states(bp.get_beliefs(bp_arrays))
+img = output.map_states[variables]
 fig, ax = plt.subplots(1, 1, figsize=(10, 10))
 ax.imshow(img)
 
--- a/examples/pmp_binary_deconvolution.py
+++ b/examples/pmp_binary_deconvolution.py
@@ -224,7 +224,7 @@ print("Time", time.time() - start)
 
 # %%
 pW = 0.25
-pS = 1e-72
+pS = 1e-80
 pX = 1e-100
 
 # Sparsity inducing priors for W and S
@@ -242,7 +242,7 @@ uX[..., 0] = (2 * X_gt - 1) * logit(pX)
 # We draw a batch of samples from the posterior in parallel by transforming `run_bp`/`get_beliefs` with `jax.vmap`
 
 # %%
-np.random.seed(seed=40)
+np.random.seed(seed=42)
 n_samples = 4
 
 start = time.time()
@@ -271,3 +271,5 @@ map_states = graph.decode_map_states(beliefs)
 
 # %%
 _ = plot_images(map_states[W].reshape(-1, feat_height, feat_width), nr=n_samples)
+
+# %%
--- a/examples/pmp_binary_deconvolution.py
+++ b/examples/pmp_binary_deconvolution.py
@@ -195,7 +195,7 @@ OR_factor_group = logical.ORFactorGroup(variables_for_ORFactors)
 fg.add_factors(OR_factor_group)
 print("Time", time.time() - start)
 
-for factor_type, factor_groups in fg._factor_types_to_groups.items():
+for factor_type, factor_groups in fg.factor_groups.items():
     if len(factor_groups) > 0:
         assert len(factor_groups) == 1
         print(f"The factor graph contains {factor_groups[0].num_factors} {factor_type}")
--- a/examples/pmp_binary_deconvolution.py
+++ b/examples/pmp_binary_deconvolution.py
@@ -107,13 +107,7 @@ _ = plot_images(W_gt[0], nr=1)
 #
 # See Section 5.6 of the [PMP paper](https://proceedings.neurips.cc/paper/2021/hash/07b1c04a30f798b5506c1ec5acfb9031-Abstract.html) for more details.
 
-import imp
-
 # %%
-from pgmax.fg import graph
-
-imp.reload(graph)
-
 # The dimensions of W used for the generation of X were (4, 5, 5) but we set them to (5, 6, 6)
 # to simulate a more realistic scenario in which we do not know their ground truth values
 n_feat, feat_height, feat_width = 5, 6, 6
@@ -183,7 +177,7 @@ for idx_img in tqdm(range(n_images)):
 
                             X_var = X[idx_img, idx_chan, idx_img_height, idx_img_width]
                             variables_for_ORFactors_dict[X_var].append(SW_var)
-print(time.time() - start)
+print("After loop", time.time() - start)
 
 # Add ANDFactorGroup, which is computationally efficient
 fg.add_factor_group(
--- a/examples/pmp_binary_deconvolution.py
+++ b/examples/pmp_binary_deconvolution.py
@@ -106,6 +106,8 @@ _ = plot_images(W_gt[0], nr=1)
 #  - a second set of ORFactors, which maps SW to X and model (binary) features overlapping.
 #
 # See Section 5.6 of the [PMP paper](https://proceedings.neurips.cc/paper/2021/hash/07b1c04a30f798b5506c1ec5acfb9031-Abstract.html) for more details.
+#
+# import imp
 
 import imp
 
@@ -150,7 +152,7 @@ print("Time", time.time() - start)
 start = time.time()
 # Factor graph
 fg = graph.FactorGraph(variables=[S, W, SW, X])
-print("x", time.time() - start)
+print(time.time() - start)
 
 # Define the ANDFactors
 variable_names_for_ANDFactors = []
@@ -232,7 +234,7 @@ print("Time", time.time() - start)
 
 # %%
 pW = 0.25
-pS = 1e-100
+pS = 1e-70
 pX = 1e-100
 
 # Sparsity inducing priors for W and S
@@ -248,7 +250,35 @@ uX[..., 0] = (2 * X_gt - 1) * logit(pX)
 
 # %%
 np.random.seed(seed=40)
-n_samples = 1
+
+start = time.time()
+bp_arrays = bp.init(
+    evidence_updates={
+        S: uS + np.random.gumbel(size=uS.shape),
+        W: uW + np.random.gumbel(size=uW.shape),
+        SW: np.zeros(SW.shape),
+        X: uX + np.zeros(shape=uX.shape),
+    },
+)
+print("Time", time.time() - start)
+bp_arrays = bp.run_bp(bp_arrays, num_iters=100, damping=0.5)
+print("Time", time.time() - start)
+output = bp.get_bp_output(bp_arrays)
+print("Time", time.time() - start)
+
+# %%
+_ = plot_images(output.map_states[W].reshape(-1, feat_height, feat_width), nr=1)
+
+# %%
+
+# %%
+
+# %% [markdown]
+# We draw a batch of samples from the posterior in parallel by transforming `run_bp`/`get_beliefs` with `jax.vmap`
+
+# %%
+np.random.seed(seed=40)
+n_samples = 4
 
 start = time.time()
 bp_arrays = jax.vmap(bp.init, in_axes=0, out_axes=0)(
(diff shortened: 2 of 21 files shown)


antoine-dedieu:
@StannisZhou I have pushed a first sketch

Here are some observation:
- with the way variable names are created in `NDVariableArray , we do not need to iterate over the indices as we previously were doing (plus we do not create the variables arrays)
- we do not have the `CompositeVariableGroup` class anymore: the variables of a FactorGraph are then represented as a dict where the values are the `NDVariableArray` and the keys are the hash of these. Thus, all the operations which refer to `VariableGroups` (eg `evidence_updates`, `beliefs` etc...) need to access to the hash of the NDVariableArray. I have therefore created a HashableDict class to help working with dictionnaries where the keys are the hash without them. 
- I have created a class BeliefPropagationOutputs to represent the BP output
- the following examples run fine: `RBM, PMP, test_or, test_wiring`

(Personal note: we cannot use a set to represent the variables in the graph as we will get `TypeError: '<' not supported between instances of 'NDVariableArray' and 'NDVariableArray'`  in update_evidence)

Here are some of the thoughts:
- should we delete VariableDict and VariableGroup? we would just have NDVariableArray which is probably enough
- creating the variable_names_for_factors  is currently slow  as it accesses the individual arrays. I am proposing a numba kernel under the hood to speed it up, see ...
- can the same variable be involved in 2 factors without having the factors touching at each other?

pgmax/groups/variables.py
```diff
@@ -22,27 +24,62 @@ class NDVariableArray(groups.VariableGroup):
...
-            Union[int, Tuple[int, ...]], nodes.Variable
-        ] = collections.OrderedDict()
-        for name in itertools.product(*[list(range(k)) for k in self.shape]):
-            if len(name) == 1:
-                names_to_variables[name[0]] = nodes.Variable(self.num_states)
-            else:
-                names_to_variables[name] = nodes.Variable(self.num_states)
+        vars_names = self.variable_names.flatten()
+        vars_num_states = self.num_states.flatten()
+        return list(zip(vars_names, vars_num_states))
```
StannisZhou, line 71: This might be slow (since it looks like a loop)? Do we really need all of `variables`, `variable_names` and `num_states` separately?

StannisZhou: For example is there anything wrong to just represent `variables` as a numpy array (so that `variable_names` is just `variables[:, 0]` and `num_states` is just `variables[:, 1]`? Is the issue that we want to keep individual variables as tuples?

pgmax/groups/variables.py
```diff
@@ -107,28 +147,32 @@ class VariableDict(groups.VariableGroup):
...
         """
-        names_to_variables: OrderedDict[
-            Tuple[Any, ...], nodes.Variable
-        ] = collections.OrderedDict()
-        for name in self.variable_names:
-            names_to_variables[name] = nodes.Variable(self.num_states)
+        return list(zip(self.variable_names, self.num_states))
 
-        return names_to_variables
+    def __getitem__(self, val):
```
StannisZhou, line 169: Type hints

pgmax/fg/groups.py
```diff
@@ -33,96 +28,32 @@ class VariableGroup:
...
-    def __getitem__(self, name: Hashable) -> nodes.Variable:
-        """This function is a typing overload and is overwritten by the implemented __getitem__"""
-
-    @typing.overload
-    def __getitem__(self, name: List) -> List[nodes.Variable]:
-        """This function is a typing overload and is overwritten by the implemented __getitem__"""
-
-    def __getitem__(self, name):
-        """Given a name, retrieve the associated Variable.
+    def __getitem__(self, val):
```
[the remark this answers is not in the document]
StannisZhou: I see why this is not implemented now. But ideally we should still have type hints at the various places

pgmax/groups/variables.py
```diff
@@ -198,10 +242,10 @@ def unflatten(
 
         start = 0
         data = {}
-        for name in self.variable_names:
+        for name in self.variables:
```
StannisZhou, line 245: We should change the `name` here to `variable`

pgmax/groups/variables.py
```diff
@@ -146,18 +190,18 @@ def flatten(
                 (2) data is not of the correct shape
         """
         for name in data:
-            if name not in self._names_to_variables:
+            if name not in self.variables:
```
StannisZhou, line 193: `name` -> `variable`

pgmax/fg/graph.py
```diff
@@ -53,48 +54,48 @@ class FactorGraph:
...
-    ]
+    variables: Union[groups.VariableGroup, Sequence[groups.VariableGroup]]
 
     def __post_init__(self):
-        if isinstance(self.variables, groups.VariableGroup):
-            self._variable_group = self.variables
+        import time
+
+        start = time.time()
+        if isinstance(self.variables, (vgroup.NDVariableArray, vgroup.VariableDict)):
```
StannisZhou, line 63: Why isn't this just `isinstance(self.variables, groups.VariableGroup)`? If you do it this way when we add new variable groups in the future we would need to change this line

pgmax/fg/graph.py
```diff
@@ -53,48 +54,48 @@ class FactorGraph:
...
+            self._variable_groups = [self.variables]
```
StannisZhou, line 64: The name `_variable_groups` is confusing after the refactor. We can probably just update `self.variables` (`FactorGraph` is not frozen and we can directly update) so that it's always called variables (and you can get rid of your `else`)

pgmax/fg/graph.py
```diff
@@ -108,16 +109,16 @@ def __hash__(self) -> int:
 
     def add_factor(
         self,
-        variable_names: List,
+        variables: List[Tuple],
```
StannisZhou, line 112:
```suggestion
        variables: List[Tuple[int, int]],
```

pgmax/fg/graph.py
```diff
@@ -130,21 +131,20 @@ def add_factor(
...
-            variable_names_for_factors=[variable_names],
+            variables_for_factors=[variables],
             factor_configs=factor_configs,
             log_potentials=log_potentials,
         )
         self._register_factor_group(factor_group, name)
 
     def add_factor_by_type(
-        self, variable_names: List, factor_type: type, *args, **kwargs
+        self, variables: List[int], factor_type: type, *args, **kwargs
```
StannisZhou, line 141: `List[Tuple[int, int]]`?

pgmax/fg/graph.py
```diff
@@ -108,16 +109,16 @@ def __hash__(self) -> int:
 
     def add_factor(
```
StannisZhou, line 110: I just realized that after the refactor, we no longer need to implement `add_factor`, `add_factor_by_type` and `add_factor_group` this way. Instead, we can construct factors/factor groups outside the factor graph and have a single generic `add_factor` function which takes as input the constructed factors/factor groups. This is a somewhat big change (but hopefully a simple one as you just need to delete lots of code), and it should make everything much more modular/intuitive.

pgmax/fg/graph.py
```diff
@@ -130,21 +131,20 @@ def add_factor(
...
         factor_group = EnumerationFactorGroup(
-            self._variable_group,
-            variable_names_for_factors=[variable_names],
+            variables_for_factors=[variables],
             factor_configs=factor_configs,
             log_potentials=log_potentials,
         )
         self._register_factor_group(factor_group, name)
 
     def add_factor_by_type(
```
StannisZhou, line 140: Can now be removed

pgmax/fg/graph.py
```diff
@@ -183,7 +181,7 @@ def add_factor_group(self, factory: Callable, *args, **kwargs) -> None:
                 for specifying the name of a named factor group.
         """
         name = kwargs.pop("name", None)
-        factor_group = factory(self._variable_group, *args, **kwargs)
+        factor_group = factory(*args, **kwargs)
```
StannisZhou, line 184: This function can now be removed

pgmax/fg/graph.py
```diff
@@ -422,12 +417,14 @@ class FactorGraphState:
     wiring: OrderedDict[type, nodes.Wiring]
 
     def __post_init__(self):
-        for field in self.__dataclass_fields__:
-            if isinstance(getattr(self, field), np.ndarray):
-                getattr(self, field).flags.writeable = False
+        for this_field in self.__dataclass_fields__:
```
StannisZhou, line 420: Why this change? Seems strictly worse

pgmax/fg/graph.py
```diff
@@ -841,9 +823,9 @@ class BPArrays:
     evidence: Union[np.ndarray, jnp.ndarray]
 
     def __post_init__(self):
-        for field in self.__dataclass_fields__:
-            if isinstance(getattr(self, field), np.ndarray):
-                getattr(self, field).flags.writeable = False
+        for this_field in self.__dataclass_fields__:
```
StannisZhou, line 826: Why change?

pgmax/fg/graph.py
```diff
@@ -1102,15 +1152,18 @@ def decode_map_states(beliefs: Any) -> Any:
...
-    )
+
+    @jax.jit
+    def _decode_map_states(beliefs) -> Any:
+        return jax.tree_util.tree_map(lambda x: jnp.argmax(x, axis=-1), beliefs)
+
+    map_states = {}
+    for variable_group, vgroup_beliefs in beliefs.items():
+        map_states[variable_group] = _decode_map_states(vgroup_beliefs)
     return map_states
```
StannisZhou, line 1163: Why is this necessary? `beliefs` should be a pytree and `tree_map` should be directly applicable. If that's not the case we should figure out why instead of making this hack

pgmax/fg/graph.py
```diff
@@ -1119,8 +1172,15 @@ def get_marginals(beliefs: Any) -> Any:
...
+    def _get_marginals(beliefs) -> Any:
+        return jax.tree_util.tree_map(
+            lambda x: jnp.exp(x - logsumexp(x, axis=-1, keepdims=True)),
+            beliefs,
+        )
+
+    marginals = {}
+    for variable_group, vgroup_beliefs in beliefs.items():
+        marginals[variable_group] = _get_marginals(vgroup_beliefs)
     return marginals
```
StannisZhou, line 1186: Why this change?

StannisZhou (commented):
Left some comments, but looks pretty good on a high level.

One major thing (something I didn't realize before) is that it looks like with the refactors we no longer need to implement the current `add_factor`/`add_factor_by_type`/`add_factor_group`. Instead, we can instantiate those outside the factor graph and have a generic `add_factor` that takes as input the constructed factors/factor groups directly. This is the benefit of using a more intuitive way to represent variables (instead of relying on names like before which requires access to a variable_group object that's only available in a factor graph).

pgmax/fg/groups.py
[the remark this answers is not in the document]
antoine-dedieu: Thanks for the tip: I don't have a mypy problem for now.

pgmax/groups/variables.py
```diff
@@ -22,27 +24,62 @@ class NDVariableArray(groups.VariableGroup):
             the notion of a NumPy ndarray shape)
     """
 
-    num_states: int
     shape: Tuple[int, ...]
+    num_states: np.ndarray
```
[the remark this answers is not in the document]
antoine-dedieu: I agree we should do that. However if we have `num_states: Union[int, np.array]` mypy does not realize that after `__postinit__`, `num_states` is always an array

pgmax/fg/graph.py
antoine-dedieu: this is because I had imported field from dataclasses. We do not need it anymore so i am happy to remove!

pgmax/groups/enumeration.py
```diff
@@ -109,18 +107,17 @@ def flatten(self, data: Union[np.ndarray, jnp.ndarray]) -> jnp.ndarray:
...
+        factor_edges_num_states = sum(
+            [variable[1] for variable in self.variables_for_factors[0]]
+        )
         if (
             data.shape != (num_factors, self.factor_configs.shape[0])
-            and data.shape
-            != (
-                num_factors,
-                np.sum(self.factors[0].edges_num_states),
-            )
```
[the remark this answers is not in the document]
antoine-dedieu: that's my understanding yes

antoine-dedieu:
Thanks for your detailed review @StannisZhou!
I have updated the process of adding factors, we should now have almost all the pieces we need

pgmax/groups/variables.py
antoine-dedieu: I have played with this and it turns out that `list(zip(vars_names, vars_num_states))` is the fastest way to keep individual variables as tuples

StannisZhou (commented):
Made another round of comments. Major issues are:

1. We should consider alternative ways for implementing variable group hashes. And checking for duplicate variables is probably not necessary if we implement our hash right.
2. Hash for `VariableDict` seems wrong
3. Remove `name` for factors/factor groups, and unify `add_factor`/`add_factor_group`
4. Properly implement flatten/unflatten for variable number of states in `NDVariableArray` (quite straightforward)

pgmax/fg/groups.py
```diff
@@ -432,6 +157,22 @@ def _variables_to_factors(self) -> Mapping[FrozenSet, nodes.Factor]:
         """
         return self._get_variables_to_factors()
 
+    @cached_property
+    def total_num_states(self) -> int:
```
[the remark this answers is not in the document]
antoine-dedieu: FactorGraph.compute_offsets

pgmax/groups/enumeration.py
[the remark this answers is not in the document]
antoine-dedieu: yes

antoine-dedieu:
@StannisZhou the PR is ready to be reviewed!

Here is an approximate timing comparisons from before / after on two examples

RBM after:
- variables +fg: 0.001s
- creating variables_to_factors 1.5s
- creating factors: 1.2s (.3s of factor_edges_num_states / .6s of checking size of log potentials)
- adding factors: 1s (.6s of test when registering)
- wiring: 1s

RBM before:
- variables + fg < 0.01s
- adding factors: 4s
- wiring: 1s


PMP after:
- variables: 0.001s
- fg: .8s (mainly vars_to_starts)
- variables_to_factors 4s
- factors: 1.5s
- wiring: 1.5s

PMP before:
- variables: 2.6s
- fg: 4.3s
- factors: 3.5s (0.9s for variable_names_to_factors)
- wiring: 1.2s

examples/ising_model.py
```diff
@@ -29,21 +29,21 @@
...
+
+variables_for_factors = []
 for ii in range(50):
     for jj in range(50):
         kk = (ii + 1) % 50
         ll = (jj + 1) % 50
-        variable_names_for_factors.append([(ii, jj), (kk, jj)])
-        variable_names_for_factors.append([(ii, jj), (ii, ll)])
+        variables_for_factors.append([variables[ii, jj], variables[kk, jj]])
+        variables_for_factors.append([variables[ii, jj], variables[kk, ll]])
```
StannisZhou, line 40: Typo? Should be `variables[ii, ll]`?

examples/ising_model.py
```diff
@@ -52,13 +52,16 @@
 bp = graph.BP(fg.bp_state, temperature=0)
 
 # %%
+# TODO: check\ time for before BP vs time for BP
+# TODO: time PGMAX vs PMP
```
StannisZhou, line 56: Remove unnecessary comments

examples/pmp_binary_deconvolution.py
```diff
@@ -116,6 +116,9 @@ def plot_images(images, display=True, nr=None):
 s_height = im_height - feat_height + 1
 s_width = im_width - feat_width + 1
 
+import time
```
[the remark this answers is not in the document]
StannisZhou: Are we going to keep these timing statements?

examples/pmp_binary_deconvolution.py
```diff
@@ -132,17 +135,20 @@ def plot_images(images, display=True, nr=None):
 
 # Binary images obtained by convolution
 X = vgroup.NDVariableArray(num_states=2, shape=X_gt.shape)
+print("Time", time.time() - start)
 
 # %% [markdown]
-# For computation efficiency, we add large FactorGroups via `fg.add_factor_group` instead of adding individual Factors
+# For computation efficiency, we add large FactorGroups via `fg.add_factors` instead of adding individual Factors
```
StannisZhou, line 141:
```suggestion
# For computation efficiency, we construct large FactorGroups instead of individual Factors
```

examples/pmp_binary_deconvolution.py
```diff
@@ -152,54 +158,44 @@ def plot_images(images, display=True, nr=None):
...
-fg.add_factor_group(
-    factory=logical.ORFactorGroup,
-    variable_names_for_factors=variable_names_for_ORFactors,
-)
+OR_factor_group = logical.ORFactorGroup(variables_for_ORFactors)
+fg.add_factors(OR_factor_group)
+print("Time", time.time() - start)
 
-for factor_type, factor_groups in fg.factor_groups.items():
+for factor_type, factor_groups in fg._factor_types_to_groups.items():
```
StannisZhou, line 198: We should avoid directly calling attributes starting with an underscore

examples/rcn.py
```diff
@@ -271,18 +272,20 @@ def valid_configs(r: int, hps: int, vps: int) -> np.ndarray:
...
 
     for e in edge:
         i1, i2, r = e
-        fg.add_factor(
-            [(idx, i1), (idx, i2)],
-            valid_configs_list[r],
+        factor = EnumerationFactor(
+            variables=[variables_all_models[idx][i1], variables_all_models[idx][i2]],
+            factor_configs=valid_configs_list[r],
+            log_potentials=np.zeros(valid_configs_list[r].shape[0]),
```
StannisZhou, line 286: This should automatically default to an all 0 array if omitted

pgmax/fg/groups.py
```diff
@@ -151,255 +96,60 @@ def unflatten(self, flat_data: Union[np.ndarray, jnp.ndarray]) -> Any:
...
+                for variables_for_factor in self.variables_for_factors
+            ]
+        )
+        object.__setattr__(self, "factor_sizes", factor_sizes)
+
+        factor_edges_num_states = np.empty(shape=(self.factor_sizes.sum(),), dtype=int)
+        _compile_edges_num_states_numba(
+            factor_edges_num_states, self.variables_for_factors
         )
+        object.__setattr__(self, "factor_edges_num_states", factor_edges_num_states)
```
StannisZhou, line 143: These are pre-computed to speed things up? Can we make then cached properties instead of attributes?

pgmax/fg/groups.py
```diff
@@ -538,6 +287,10 @@ def __post_init__(self):
             if not hasattr(self, key):
                 object.__setattr__(self, key, getattr(self.factor, key))
 
+        object.__setattr__(
+            self, "log_potentials", getattr(self.factor, "log_potentials")
+        )
```
StannisZhou, line 292: Why didn't we have this before?

pgmax/fg/groups.py
```diff
@@ -575,3 +326,14 @@ def unflatten(self, flat_data: Union[np.ndarray, jnp.ndarray]) -> Any:
         raise NotImplementedError(
             "SingleFactorGroup does not support vectorized factor operations."
         )
+
+
+nb.jit(parallel=False, cache=True, fastmath=True, nopython=False)
+
+
```
StannisZhou, line 333: Formatting seems strange here...Did you misplace the decorator?

pgmax/groups/enumeration.py
```diff
@@ -231,36 +224,37 @@ def __post_init__(self):
...
-                f"Expected log_potential_matrix for {len(self.variable_names_for_factors)} factors. "
+                f"Expected log_potential_matrix for {len(self.variables_for_factors)} factors. "
                 f"Got log_potential_matrix for {log_potential_matrix.shape[0]} factors."
             )
 
-        for fac_list in self.variable_names_for_factors:
-            if len(fac_list) != 2:
+        import time
+
+        start = time.time()
```
StannisZhou, line 236: Remove

pgmax/groups/enumeration.py
```diff
@@ -231,36 +224,37 @@ def __post_init__(self):
...
                 raise ValueError(
-                    f"The specified pairwise factor {fac_list} (with "
-                    f"{(self.variable_group[fac_list[0]].num_states, self.variable_group[fac_list[1]].num_states)} "
+                    f"The specified pairwise factor {variables_for_factor} (with {factor_num_configs}"
                     f"configurations) does not match the specified log_potential_matrix "
-                    f"(with {log_potential_matrix.shape[-2:]} configurations)."
+                    f"(with {log_potential_shape} configurations)."
                 )
-        object.__setattr__(self, "log_potential_matrix", log_potential_matrix)
+            object.__setattr__(self, "log_potential_matrix", log_potential_matrix)
```
StannisZhou, line 255: Wrong indentation, you are setting this many times

pgmax/groups/variables.py
```diff
@@ -1,73 +1,136 @@
...
+            )
 
-    def _get_names_to_variables(
-        self,
-    ) -> OrderedDict[Union[int, Tuple[int, ...]], nodes.Variable]:
-        """Function that generates a dictionary mapping names to variables.
+        # Only compute the hash once, which is guaranteed to be an int64
+        this_id = id(self) % 2**32
+        this_hash = this_id * int(MAX_SIZE)
+        assert this_hash < 2**63
```
StannisZhou, line 54: Not sure about this assert...Given you are doing the `%`, I assume this should never fail? 

A separate question is, although there won't be collisions for `id(self)`, is it possible that there will be collisions for `id(self) % 2**32`? Why are we doing the `%` here?

pgmax/groups/variables.py
```diff
@@ -1,73 +1,136 @@
...
+        object.__setattr__(self, "this_hash", this_hash)
```
StannisZhou, line 55: Why can't we have this in the parent class?

pgmax/groups/variables.py
```diff
@@ -1,73 +1,136 @@
...
+        ):
+            assert isinstance(self.num_states, np.ndarray)
+            vars_names = self.variable_names[val].flatten()
+            vars_num_states = self.num_states[val].flatten()
+            return list(zip(vars_names, vars_num_states))
+
+        return (self.variable_names[val], self.num_states[val])
+
+    @cached_property
+    def variables(self) -> List[Tuple]:
```
StannisZhou, line 84: `Tuple[int, int]`

pgmax/groups/variables.py
```diff
@@ -106,29 +176,57 @@ class VariableDict(groups.VariableGroup):
...
 
-    def _get_names_to_variables(self) -> OrderedDict[Tuple[int, ...], nodes.Variable]:
-        """Function that generates a dictionary mapping names to variables.
+        # Only compute the hash once
+        object.__setattr__(self, "this_hash", id(self))
+
+        hash_and_names = tuple(
+            (self.__hash__(), var_name) for var_name in self.variable_names
+        )
+        object.__setattr__(self, "variable_names", hash_and_names)
```
StannisZhou, line 198: Make this into a cached property

pgmax/groups/variables.py
```diff
@@ -106,29 +176,57 @@ class VariableDict(groups.VariableGroup):
...
         num_states: The size of the variables in this variable group
-        variable_names: A tuple of all names of the variables in this variable group
-
+        variable_names: A tuple of all names of the variables in this variable group.
+            Note that we overwrite variable_names to add the hash of the VariableDict
     """
 
-    num_states: int
     variable_names: Tuple[Any, ...]
+    num_states: int
```
StannisZhou, line 184: Any reason not to support variable number of states? Should be simple and the same as `NDVariableArray`

pgmax/groups/variables.py
```diff
@@ -106,29 +176,57 @@ class VariableDict(groups.VariableGroup):
...
+
+    def __post_init__(self):
+        num_states = np.full(
+            (len(self.variable_names),), fill_value=self.num_states, dtype=np.int32
+        )
+        object.__setattr__(self, "num_states", num_states)
```
StannisZhou, line 190: Support variable number of states. Otherwise no reason to make this an ndarray

StannisZhou (commented):
Partial review. Will finish the rest tomorrow...

pgmax/groups/enumeration.py
antoine-dedieu: will do before merging

pgmax/groups/variables.py
antoine-dedieu: to make sure that the hash is int64

pgmax/groups/variables.py
antoine-dedieu: that's not true in the case of VariableDict

examples/ising_model.py
antoine-dedieu: will do before merging

pgmax/groups/variables.py
antoine-dedieu: no: we are overwritting `variable_names` to add the hash

pgmax/groups/enumeration.py
antoine-dedieu: thanks :)

pgmax/groups/variables.py
wlehrach: Out of curiosity, why don't we use the nicer `tuple[int, int]` form?

StannisZhou: We can make a global switch! We started with `Tuple` so for consistency has been using `Tuple`

pgmax/fg/groups.py
antoine-dedieu: because for SingleFactorGroup, we were calling `factor.compile_wiring`

examples/rcn.py
antoine-dedieu: This is true for `EnumerationFactorGroup`, not for `EnumerationFactor`

pgmax/fg/graph.py
```diff
@@ -45,181 +43,86 @@ class FactorGraph:
...
-        name = kwargs.pop("name", None)
-        factor_group = factory(self._variable_group, *args, **kwargs)
-        self._register_factor_group(factor_group, name)
-
-    def _register_factor_group(
-        self, factor_group: groups.FactorGroup, name: Optional[str] = None
-    ) -> None:
-        """Register a factor group to the factor graph, by updating the factor graph state.
+        if factor is None and factor_group is None:
+            raise ValueError("A Factor or a FactorGroup is required")
```
StannisZhou, line 104: This is pretty unintuitive. In theory we can easy handle this case (just process both). But there is also no reason to restrict to handling a single factor or a single factor group.

More generally, I think it's better if we just have a single input called factors, which can be a single factor, a single factor group, or a sequence of factor/factor groups. Inside the function we have a few if/else statements to handle various cases and add all the factors.

pgmax/fg/graph.py
```diff
@@ -45,181 +43,86 @@ class FactorGraph:
...
-        """
-        if name in self._named_factor_groups:
-            raise ValueError(
-                f"A factor group with the name {name} already exists. Please choose a different name!"
+        if factor is not None:
+            factor_group = groups.SingleFactorGroup(
+                variables_for_factors=[factor.variables],
+                factor=factor,
             )
+        assert factor_group is not None
```
StannisZhou, line 114: Let's not do assert here but instead make sure the logic is solid and `factor_group` can never be `None`

pgmax/fg/graph.py
```diff
@@ -337,20 +240,28 @@ def factors(self) -> OrderedDict[Type, Tuple[nodes.Factor, ...]]:
...
+    def factor_groups(self) -> Tuple[groups.FactorGroup, ...]:
         """Tuple of factor groups in the factor graph"""
-        return self._factor_types_to_groups
+        return tuple(
+            [
+                factor_group
+                for factor_groups_per_type in self._factor_types_to_groups.values()
+                for factor_group in factor_groups_per_type
+            ]
+        )
```
StannisZhou, line 264: Why this change? To be used in the hash? Looks equivalent to the previous implementation (i.e. it doesn't seem like we achieved anything with this change and the change in the hash). This also looks to be the reason you have to access a private attribute in one of the notebooks

pgmax/fg/graph.py
```diff
@@ -396,25 +308,25 @@ class FactorGraphState:
...
     """
 
-    variable_group: groups.VariableGroup
-    vars_to_starts: Mapping[nodes.Variable, int]
+    variable_groups: Sequence[groups.VariableGroup]
+    vars_to_starts: Mapping[Tuple[Any, int], int]
     num_var_states: int
     total_factor_num_states: int
-    named_factor_groups: Mapping[Hashable, groups.FactorGroup]
+    factor_groups: Tuple[groups.FactorGroup, ...]
```
StannisZhou, line 329: Looks like this `factor_groups` is only used to check whether a given `factor_group` exists. If so it's redundant (you can just check whether a given `factor_group` is `in factor_group_to_potential_starts`)

pgmax/fg/graph.py
```diff
@@ -482,24 +394,21 @@ def update_log_potentials(
         (1) Provided log_potentials shape does not match the expected log_potentials shape.
         (2) Provided name is not valid for log_potentials updates.
     """
-    for name in updates:
-        data = updates[name]
-        if name in fg_state.named_factor_groups:
-            factor_group = fg_state.named_factor_groups[name]
-
+    for factor_group, data in updates.items():
```
StannisZhou, line 397: This means there is no way to pass in updates for individual factors. I think this is a case that is currently not covered (solution is to use single factor as keys when adding single factors)

pgmax/groups/variables.py
StannisZhou: OK I missed that...That naming is confusing, though. Can we change the `variable_names` containing the hash to `variable_identifiers`? Those are not really names in the first place

pgmax/groups/variables.py
```diff
@@ -105,30 +170,67 @@ class VariableDict(groups.VariableGroup):
...
+
+        Returns:
+            The queried variable
+        """
+        assert isinstance(self.num_states, np.ndarray)
+        if (self.__hash__(), val) not in self.variable_names:
+            raise ValueError(f"Variable {val} is not in VariableDict")
+
+        idx = self.variable_names.index((self.__hash__(), val))
+        return ((self.__hash__(), val), self.num_states[idx])
```
StannisZhou, line 230: For consistency it would be best if we can make all variables `Tuple[int, int]`. I'm actually a bit surprised that mypy is passing right now. There are places where you assume a variable is always represented as `Tuple[int, int]` but they are not and for some reason mypy still thinks things are consistent

pgmax/fg/groups.py
```diff
@@ -575,3 +335,12 @@ def unflatten(self, flat_data: Union[np.ndarray, jnp.ndarray]) -> Any:
         raise NotImplementedError(
             "SingleFactorGroup does not support vectorized factor operations."
         )
+
+
+# @nb.jit(parallel=False, cache=True, fastmath=True)
```
StannisZhou, line 340: Uncomment?

StannisZhou (commented):
More comments

pgmax/fg/graph.py
antoine-dedieu: Correct, we currently do not provide this option. In factor we do not compute the `factor_to_potentials_starts` indices

examples/pmp_binary_deconvolution.py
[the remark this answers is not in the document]
antoine-dedieu: No they will be removed before merging. Good to keep them until then though, to make sure we do not accidentally slow things down

pgmax/fg/groups.py
```diff
@@ -419,9 +174,40 @@ def __getitem__(self, variables: Union[Sequence, Collection]) -> Any:
...
+    def factor_edges_num_states(self) -> np.ndarray:
+        """Computes an array concatenating the number of states for the variables connected to each Factor of
+        the FactorGroup. Each variable will appear once for each Factor it connects to.
+
+        Returns:
+            Array with the the number of states for the variables connected to each Factor.
+        """
+        factor_edges_num_states = np.empty(shape=(self.factor_sizes.sum(),), dtype=int)
+
+        # TODO: create variables_for_factors as an array and move this to numba
```
StannisZhou, line 203: Track as an issue instead of having TODOs in the code?

pgmax/fg/graph.py
```diff
@@ -45,56 +43,43 @@ class FactorGraph:
...
         # See FactorGraphState docstrings for documentation on the following fields
-        self._num_var_states = vars_num_states_cumsum[-1]
-        self._vars_to_starts = MappingProxyType(
-            {
-                variable: vars_num_states_cumsum[vv]
-                for vv, variable in enumerate(self._variable_group.variables)
-            }
-        )
-        self._named_factor_groups: Dict[Hashable, groups.FactorGroup] = {}
+        # TODO: vars_to_starts does not have to be a dict
```
StannisZhou, line 68: Create an issue to track instead of leaving TODOs in code

pgmax/fg/graph.py
```diff
@@ -106,120 +91,46 @@ def __hash__(self) -> int:
...
         """
-        if name in self._named_factor_groups:
-            raise ValueError(
-                f"A factor group with the name {name} already exists. Please choose a different name!"
+        if isinstance(factors, list):
+            for factor in factors:
+                self.add_factors(factor)
+            return None
+
+        if isinstance(factors, groups.FactorGroup):
```
StannisZhou, line 116: Why not just use `elif`? The return seems a bit awkward

pgmax/fg/groups.py
```diff
@@ -25,104 +22,76 @@
...
-
-    @typing.overload
-    def __getitem__(self, name: Hashable) -> nodes.Variable:
-        """This function is a typing overload and is overwritten by the implemented __getitem__"""
+    def __post_init__(self):
+        # Only compute the hash once, which is guaranteed to be an int64
+        this_id = id(self) % 2**32
+        this_hash = this_id * int(MAX_SIZE)
+        assert this_hash < 2**63
+        object.__setattr__(self, "this_hash", this_hash)
```
StannisZhou, line 46: Can we change `this_hash` to something like `_hash` to emphasize that this shouldn't be publicly accessible?

pgmax/fg/groups.py
StannisZhou: This is still commented. Is this intended?

pgmax/fg/groups.py
StannisZhou: Still not sure this is needed...

StannisZhou (approved):
LGTM. Thanks for patiently addressing all the comments!

PR #144, opened 2022-05-19
Hotfix

antoine-dedieu:
Fixing two minor issues with variable number of states


StannisZhou (commented):
Could you also add a simple unit test that would catch this earlier, and bump up the version for another release?

StannisZhou (approved):
LGTM. Thanks for fixing!

PR #147, opened 2023-02-04
Porting PGMax over

antoine-dedieu:


README.md
```diff
@@ -6,6 +6,14 @@
 
 # PGMax
 
+## PGMax is now part of DeepMind 🎉!
```
StannisZhou, line 9: Better to make this top-level, above everything else, instead of part of the description

StannisZhou: Also you should make this notice bigger and stand out more. https://github.com/ylabbe/cosypose/blob/master/README.md Here is an example that I recently came across. You could adapt their html code.

antoine-dedieu: these points should be adressed in the new commit

README.md
```diff
@@ -4,8 +4,28 @@
...
+<a href="https://github.com/deepmind/PGMax">DeepMind's repository</a>.
+</h1>
+</body>
+<br>
+<br>
+
+
 # PGMax
 
+## Description
```
StannisZhou, line 27: We should also update the descriptions and get rid of the obsolete info to avoid confusing people.

One option is just remove all the old descriptions.

Another is to slightly update the old descriptions. Things I can think of:

1. Remove blog post link
2. Update citation entry
3. Maybe update installation instructions

StannisZhou (commented):
The old wording is a bit long and repetitive, so I made some suggestions.

I also feel we should update the old descriptions to avoid confusing people.

README.md
antoine-dedieu: good point: let's just remove all the descriptions

